PR #1, opened 2021-01-18
[ch51]  Create frontend codebase

kimvcp:
https://app.clubhouse.io/cobda/story/51/create-frontend-code-base

## Why?

This PR is for initializing the frontend codebase.

## Changes

- Init codebase with `npx create-react-app cobda-web --template typescript` command.

## Screenshot

See change [here](https://github.com/Cobda/cobda-web/pull/1/files).


src/App.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+export default function App() {
+  return (
+    <div> Hello World
+    </div>
+  );
+}
```
LilBank: Please add a newline.

LilBank (changes requested):

public/index.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
```
LilBank, line 3: Is it possible to create this into an individual component? Whereas in future there will be many implementations in this part.

tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
LilBank: Quick question, should we upgrade this to `es6`?

kimvcp: absolutely yes 👍

public/index.html
kimvcp: Yes, it's possible, but since we don't plan to dynamically change what is inside `<head>` , we can use the generated one as our “entry point”. Eventually, implementing them in another PR is a better idea. Isn't it? 🧐

LilBank: Okay then, lets leave it first. Will be implement in future.

public/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+	"short_name": "React App",
+	"name": "Create React App Sample",
```
Raksani: To be sure, Is this necessary to be our project name?

Raksani (commented):
Great. Everything looks fine.

Raksani:
I just curious that we supposed to use the word "code base" or "codebase". I think "codebase" is correct for me.

public/manifest.json
kimvcp: Actually, we don't need this file.

package.json
```diff
@@ -0,0 +1,43 @@
...
+    "@testing-library/jest-dom": "^5.11.4",
+    "@testing-library/react": "^11.1.0",
+    "@testing-library/user-event": "^12.1.10",
+    "@types/jest": "^26.0.15",
+    "@types/node": "^12.0.0",
+    "@types/react": "^16.9.53",
+    "@types/react-dom": "^16.9.8",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
+    "react-scripts": "4.0.1",
```
LilBank: Apologize for late question, do you think this dependency has to keep updated too? I'm not quite sure but please recheck.
```suggestion
    "react-scripts": "^4.0.1",
```

LilBank: FYI, in future since the backend code will be put in here too, we might not use `react-scripts` so you can just keep it like this first. But it's still good to check 🧐

Raksani (commented):
Everything looks fine to me.

Raksani (approved):

package.json
kimvcp: Since `react-scripts` version `4.0.1` is already the latest one, this will work perfectly fine for now and soon. However, I agreed to keep it updated for more flexibility.

package.json
```diff
@@ -1,43 +1,43 @@
...
-      "not dead",
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
-  }
+	"name": "cobda-web",
```
LilBank: Please recheck the indention here. Let's stick with 2 spaced indent

LilBank (changes requested):

package.json
kimvcp: Fixed

LilBank (approved):

PR #3, opened 2021-01-20
[ch60] As a user, I can see homepage separated into different sections

kimvcp:
https://app.clubhouse.io/cobda/story/60/as-a-user-i-can-see-homepage-separated-into-different-sections

## Why?
This **PR** is the entry point for the homepage. It separates the homepage into 3 sections which are header, main, footer. 

## Changes
- Init SCSS file structure.
- Init React Router.
- Add utility functions for implementing responsive layout.
- Add Navbar, Sidebar, Footer components.

### Before 

Screen Shot 2564-01-20 at 22 26 43

### After

Smaller than the Medium Device (MD).

Screen Shot 2564-01-20 at 22 14 26

Larger than the Medium Device (MD).

Screen Shot 2564-01-20 at 22 14 42


package.json
```diff
@@ -10,6 +10,8 @@
     "@types/node": "^12.0.0",
     "@types/react": "^16.9.53",
     "@types/react-dom": "^16.9.8",
+    "@types/react-router-dom": "^5.1.7",
+    "node-sass": "4.14.1",
```
LilBank, line 14: Quick question, does dependency needs to keep updated too?

src/styles/_functions.scss
```diff
@@ -0,0 +1,4 @@
+@function rem($size) {
+  $remSize: $size / 16;
```
LilBank: I'm not quite sure but is it possible to handle the `$size` input that it should always be in `px` unit?

src/components/Footer/styles.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/main.scss';
+
+.footer-container {
+  height: rem(32);
```
LilBank: Should we use `px` unit?
```suggestion
  height: rem(32px);
```

src/components/Footer/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import './styles.scss';
+
+export default function Footer() {
+  return (
+    <section className='footer-container'>
```
LilBank: Quick question, for html is `.tsx`, should we use double quote?
```suggestion
    <section className="footer-container">
```

src/components/Sidebar/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import './styles.scss';
+
+export default function Sidebar() {
+  return (
+    <section className='sidebar-container'>
```
LilBank: When the screen is smaller than `md`, the sidebar should stay aside footer (it should align horizontally).

src/styles/_mixins.scss
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+
+@mixin flex-center() {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+@mixin sticky() {
```
LilBank: If this `@mixin` will only be use with `navbar`, should we move all the properties to its `css` class?

src/screens/Home/index.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import Footer from '../../components/Footer';
+import Navbar from '../../components/Navbar';
+import Sidebar from '../../components/Sidebar';
+import './styles.scss';
+
+export default function Home() {
+  return (
+    <>
```
LilBank, line 9: Quick question, what is this?

src/components/Footer/index.tsx
```diff
@@ -0,0 +1,10 @@
...
+      <div>Footer</div>
+    </section>
```
LilBank: `<section>` element should have heading(`h1-h6`). Since there's no heading, could we avoid unnecessary nested `<div>`? What do you think?

Source: https://stackoverflow.com/questions/6939864/what-is-the-difference-between-section-and-div
```suggestion
    <div className='footer-container'>Footer</div>
```

LilBank (changes requested):

package.json
kimvcp: Nope, this specific sass version fits the most with other dependencies.

src/components/Footer/styles.scss
Kittinske15: I think rem is a unit already, no need to put px scaling it.

src/styles/_mixins.scss
kimvcp: As I mentioned, this PR is mainly the entry point for others. Wouldn't it be nicer to leave `@mixin` there for later implementation? What do you think?

src/screens/Home/index.tsx
kimvcp: Normally, we have to declare a "React Fragment (<>)" to ensure that all the children are grouped without adding extra nodes to the DOM.

src/components/Sidebar/index.tsx
kimvcp: Could you give some sample screenshots for more illustration?

src/components/Footer/index.tsx
kimvcp: Okay thanks 🤓

src/components/Footer/index.tsx
kimvcp: Yep, all JSX elements should be double quote instead.

src/styles/_functions.scss
kimvcp: Agreed, now it's handled.

src/components/Footer/styles.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/main.scss';
+
+.footer-container {
+  height: rem(32px);
```
Raksani, line 4: I think we should choose between `rem` or `px`.

src/components/Footer/styles.scss
```diff
@@ -0,0 +1,11 @@
...
+  background-color: $secondary;
+
+  @include flex-center();
+  @include respond-above(md) {
+    height: rem(72px);
```
Raksani, line 9: They're different units, aren't they?

src/components/Navbar/styles.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../styles/main.scss';
+
+.navbar-container {
+  height: rem(32px);
```
Raksani, line 10: Does it suppose to contain only 1 unit?

src/components/Navbar/styles.scss
```diff
@@ -0,0 +1,12 @@
...
+  background-color: $background;
+
+  @include sticky();
+  @include flex-center();
+  @include respond-above(md) {
+    height: rem(72px);
```
Raksani, line 16: It has to be only one unit `rem` or `px`. Correct me if I misunderstood.

src/components/Sidebar/index.tsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import './styles.scss';
```
Raksani, line 2: Just curious, do we need the separate stylesheet one for the `Sidebar`?
In further changes, it might contain filtering or categorizing features.

src/screens/Home/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+      <header>
+        <Navbar />
+      </header>
+      <main>
```
Raksani, line 13: Does our main area including only `Sidebar`?

src/styles/_functions.scss
```diff
@@ -0,0 +1,4 @@
+@function rem($size) {
+  $remSize: $size / 16px;
```
Raksani: Just curious, why we have to use two units?

Raksani (changes requested):

Raksani:
To me personally, I think the folder name `Screens` seems unfitting. 'Pages' is a better choice.

LilBank:
> To me personally, I think the folder name `Screens` seems unfitting. 'Pages' is a better choice.

`Screens` is the best practice and widely use. I think we should keep it that way 😆. For example:
- https://nimblehq.co/compass/development/code-conventions/react/#project-structure
- https://nimblehq.co/compass/development/code-conventions/javascript/#js-only-applications
- https://nimblehq.co/compass/development/code-conventions/css/#stylesheets-structure

src/components/Footer/styles.scss
kimvcp: Yeah, we actually choose `rem`. The code you see is an SCSS function for converting `px` to `rem`. We have `rem` as a function name and `32px` for its parameter. Do you think we should change its name or something?

src/components/Navbar/styles.scss
kimvcp: Do you mean writting `rem(32)` instead?

src/components/Navbar/styles.scss
kimvcp: Do you mean writting `rem(72)` instead?

src/screens/Home/index.tsx
kimvcp: Absolutely no, we will surely add more components in later `PR`. If you think this is confusing, I can add some comments inside.

src/styles/_functions.scss
kimvcp: Since `rem` is more flexible and scalable than `px`, we will use `rem` as our main unit.
Ref: https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css

kimvcp: Normally the design comes with `px` unit instead of `rem`. This converter will make our code more readable. For example, we could write rem(150px) instead of 9.375rem.

src/components/Sidebar/index.tsx
kimvcp: Sorry, I don't understand what you are trying to say?

src/components/Footer/styles.scss
LilBank: The name is fine actually, not a good practice if we rename it to something like `px_to_rem`.

src/components/Footer/styles.scss
LilBank: In this case, since you've a suggestion on this [LOC](https://github.com/Cobda/cobda-web/pull/3/files#r561194816), you should also move this comment to that one too. Prevent opening multiple comments with same context 😎

src/components/Navbar/styles.scss
LilBank: Same as above.

src/screens/Home/index.tsx
LilBank: We can align our understanding here so we don't have to put unnecessary comments.

src/styles/_mixins.scss
LilBank: Can you please give more examples on which component will use this `@mixin` other than `navbar`? 🤔

src/components/Sidebar/index.tsx
LilBank: Notice the sidebar

Screen Shot 2564-01-21 at 10 46 42

Now in small screen, it stay besides footer

Screen Shot 2564-01-21 at 10 47 06

kimvcp: Fixing this could possibly be a waste of time since we don't know what designs are going to be. Can we keep it still for now and adjust them later according to the designs?

src/styles/_mixins.scss
kimvcp: At first, I was thinking that it could be useful for other components inside `<header>`. But since we don't have the design yet, hence moving this to its own separate file would be a better choice.

src/components/Sidebar/index.tsx
LilBank: Okay 👌

src/components/Sidebar/index.tsx
Raksani: I think I should move this topic to the main PR.

Raksani:
@kimvcp As I mentioned about scss file, apologies for being unclear, the first time I saw I think we have only one scss file since they have the same name. Would it be better to rename `styles.scss` to be something with uniqueness like `_sidebar.scss`. I tried to find the best practice for scss architecture and I found this article seems pretty good [https://matthewelsom.com/blog/simple-scss-playbook.html](url). Anyway, I leave the decision up to you.

kimvcp:
> @kimvcp As I mentioned about scss file, apologies for being unclear, the first time I saw I think we have only one scss file since they have the same name. Would it be better to rename `styles.scss` to be something with uniqueness like `_sidebar.scss`. I tried to find the best practice for scss architecture and I found this article seems pretty good [https://matthewelsom.com/blog/simple-scss-playbook.html](url). Anyway, I leave the decision up to you.

kimvcp:
> @kimvcp As I mentioned about scss file, apologies for being unclear, the first time I saw I think we have only one scss file since they have the same name. Would it be better to rename `styles.scss` to be something with uniqueness like `_sidebar.scss`. I tried to find the best practice for scss architecture and I found this article seems pretty good [https://matthewelsom.com/blog/simple-scss-playbook.html](url). Anyway, I leave the decision up to you.

This article suggests that we should group all the styles into a separate folder right? Correct me if I am wrong.
In my experience, I prefer to have a `.scss` file for each component's folder instead. Therefore, the folder name for each component is already unique in itself, so we could have `index.tsx` and `styles.scss` inside without concerning about the uniqueness.
Can I know your opinion? @LilBank @Kittinske15

LilBank (approved):

Raksani (approved):

Kittinske15 (approved):
I have reviewed all and It's look amazing

PR #5, opened 2021-01-22
[ch68] As a user, I can see Product List screen

Raksani:
https://app.clubhouse.io/cobda/story/68/as-a-user-i-can-see-product-list-screen

## Why?

This **PR** is the entry point for the`Product List` screen. And accessible by clicking `Search` button which is located in the center of the `body` section.

## Changes
- Create Product List screen.
- Add Search button to the center of the homepage which leads to `Product List` screen.

## Screenshot(s)
**Before** 
105196496-478ec080-5b6e-11eb-8bb6-57f55272deca
**After** 
Search button is located in the center of the homepage.
Screenshot 2564-02-01 at 18 02 05

After clicking `Search` button, redirect to product-list screen.
Screenshot 2564-02-02 at 01 05 36

PR #4, opened 2021-01-22
[ch64] As a user, I can see About Us screen

Raksani:
[https://app.clubhouse.io/cobda/story/64/as-a-user-i-can-see-about-us-screen](url)

## Why?

This **PR** is the entry point for the `About Us` screen. And accessible through the `Footer` component. 

## Changes

- Create `About Us` screen.
- Add `About us` link on the `Footer`.

## Screenshot(s)

**Before**
105196496-478ec080-5b6e-11eb-8bb6-57f55272deca

**After**
About us link appears on Footer. And About Us screens was created.
Screenshot 2564-01-22 at 23 11 19

PR #6, opened 2021-01-22
[ch69] As a user, I can see Product Detail screen

Raksani:
[https://app.clubhouse.io/cobda/story/69/as-a-user-i-can-see-product-detail-screen](url)

## Why?

This **PR** is the entry point for the `Product Detail` screen. And accessible by clicking `Product Detail` button which leads to `Product List` screen.

## Changes
- Create `Product Detail` screen.
- Add `Product Detail` button in `Product List` screen which is redirecting to Product Detail screen.

## Screenshot(s)
**Before**
Screenshot 2564-01-23 at 01 42 41
**After**
Product Detail screen was created.
Screenshot 2564-02-06 at 21 31 28
Product List is clickable to Product Detail page.
Screenshot 2564-02-06 at 21 31 15


LilBank:
Screen Shot 2564-01-23 at 14 32 37
This PR should be merge into `feature/product-list-ui` isn't it? 🤔

LilBank:
Please update PR description in **Changes** section.
Source: https://github.com/Cobda/cobda-web/pull/4#issuecomment-765882629

src/components/Search/styles.scss
```diff
@@ -0,0 +1,36 @@
+@import '../../styles/main.scss';
+.search-container {
+    float: right;
+    height: 100vh;
+    overflow: auto;
+    background-color: white;
+    @include flex-center();
+    @include respond-above(md) {
+      width: rem(1140px);
```
LilBank: Reminder, this should change after a rebase from `feature/product-list-ui` 😎 (Close this comment when it changes)

src/components/Search/index.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
LilBank: Reminder, this should change after a rebase from `feature/product-list-ui` 😎 (Close this comment when it changes)

src/router/index.tsx
```diff
@@ -15,6 +17,12 @@ export default function Routes() {
         <Route exact path="/">
           <Home />
         </Route>
+        <Route exact path="/product-list">
+          <ProductList/>
+        </Route>
+        <Route exact path="/product-detail">
+          <ProductDetail/>
```
LilBank: Typo 😁
```suggestion
          <ProductDetail />
```

src/screens/ProductList/styles.scss
```diff
@@ -0,0 +1,34 @@
...
+.btn {
+  color: white;
+  text-decoration: none;
+  padding: 5px 10px;
+  border-radius: 3px;
+}
+
+.btn-product-detail {
+  @include button-bg(#5a6570);
+}
```
LilBank: See here: https://github.com/Cobda/cobda-web/pull/5#discussion_r563030936

LilBank (changes requested):
About `package-lock.json`, see here: https://github.com/Cobda/cobda-web/pull/5#pullrequestreview-574742980

Raksani:
> About `package-lock.json`, see here: [#5 (review)](https://github.com/Cobda/cobda-web/pull/5#pullrequestreview-574742980)

There's no `package-lock.json`.

src/components/ProductDetail/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+
+function ProductDetailComponent() {
+  return (
+    <section className="product-detail">
+      <div className="product-detail__content-image">
+        <div className="product-detail__image-container">
+          <div className="product-detail__image">
+            <img src="#" alt="product-image" />
+          </div>
```
LilBank: I don't think we need another `div` after the `container` though. Could have been `img` straight away 🤔

src/components/ProductList/index.tsx
```diff
@@ -1,15 +1,20 @@
...
+        <h2 className="product-list__title">
+          <Link href="/product-detail">
+            <a className="product-list__link-to-detail">Product List</a>
+          </Link>
+        </h2>
+        <p className="product-list__description">
+          This is a sample description
+        </p>
+      </header>
+    </section>
```
LilBank: I thought you were going to create a `1 x 3` `product-article`. I think it's better than creating `&__link-to-detail` class which it isn't relate to the `product-article` at all 🤔

src/components/ProductDetail/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+
+function ProductDetailComponent() {
```
LilBank: How about renaming it `ProductView`? And inside this component should contain `ProductContent`, `ProductDetail`, and `ProductInformation`? Each styling class for each component. Keep the `grid` on `ProductView` and create new 3 components.

```
const ProductView = () => {
  return (
    <section className="product-view">
      <div className="product-view__content">
        <ProductContent />
      </div>
      <div className="product-view__detail">
        <ProductDetail />
      </div>
      <div className="product-view__information">
        <ProductInformation />
      </div>
    </section>
  )
}
```

LilBank (changes requested):

src/components/ProductList/index.tsx
Raksani: First, I thought just to make it clickable in the first place as we focus on seeing the product detail screen. I can re-design it to use `article` tag instead in the next implement for the product list page.

src/components/ProductDetail/index.tsx
Raksani: After discussing, `ProductView` is a better name. But the component should be `ProductImage`, `ProductInformation`, and `ProductDescription`. Content can be anything `image` is more specific from my view.

src/components/ProductList/index.tsx
LilBank: Just a simple three `div` box should do, since I don't want to create a css class `&__link` for `product-list`.

src/components/ProductDetail/index.tsx
LilBank: ## TL;DR

Prefix `image` is for an element while `content` is for a block. A component is a block which consist of elements.

## Explanation

Yep `ProductImage` is a good name. But seems like you forgot to think about the other essentials(i.e. `slider`, `dots`, `button--arrow`). If we name the component `ProductImage`, we have to name those elements as `.product-image__slider` or `.product-image__owl-dots`. This isn't meaningful enough and violates **BEM** convention. On the other hand, `.product-content__image`/`.product-content__button--arrow`/`.product-content__slider`.

What do you think?

Raksani: As you mentioned, content as a `block` why don't we use `ProductContent` instead of `ProductView` and `ProductView` for viewing image. TBH, I'd like something more specific than just `content` because those 3 elements could be named 'content' as well as I have done before. Anyway, `ProductDetail` is used for the `page` name already we should avoid this. How about this?
```
const ProductContent = () => {
  return (
    <section className="product-content">
      <div className="product-content__view">
        <ProductView />
      </div>
      <div className="product-content__information">
        <ProductInformation />
      </div>
      <div className="product-content__description">
        <ProductDescription />
      </div>
    </section>
  )
}
```

src/components/ProductList/index.tsx
Raksani: Alright.

src/components/ProductDetail/index.tsx
Raksani: Anyone would like to bring some ideas? @kimvcp @Kittinske15?

LilBank: Firstly, to discrimate details and information: https://www.quora.com/What-is-the-difference-between-details-and-information
Let's keep the word **description** for `<p>` and use **information** instead.

Next, naming a component `ProductDetail` is not going to break the `prodct-detail.tsx` since the component were importing is `Product...`.

Furthermore, I agree with using `ProductContent` and for the image block lets use `ProductMedia`. Inspiration: https://stockx.com (please kindly have a look).

Lastly, how about `ProductDetail` to `ProductView` in `pages`?

What do you think?

Kittinske15: Agree with changing ``ProductDetail`` to ``ProductView`` in pages and also using ``product-content`` is more meaningful than ``product-view`` for my opinion.

kimvcp: I personally like how `stockx` name their components. I'm pretty okay if we decide to follow them.

Raksani: Okay then, I'll keep with this version --> https://github.com/Cobda/cobda-web/pull/6#discussion_r570333725

src/components/ProductView/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+      <div className="product-view__content">
+        <ProductContent />
+      </div>
+      <div className="product-view__detail">
+        <ProductDetail />
+      </div>
+      <div className="product-view__information">
+        <ProductInformation />
+      </div>
+    </section>
```
kimvcp: What if we move these block `(product-view_...)` into its own component instead. Do you think it looks cleaner?
```suggestion
    <section className="product-view">
      <ProductContent />
      <ProductDetail />
      <ProductInformation />
    </section>
```

kimvcp:
1. Please use `Pascal` case for naming all `.tsx` file except those under `/pages` which should be the `Kebab` case.
[Example here](https://medium.com/better-programming/string-case-styles-camel-pascal-snake-and-kebab-case-981407998841)
2. `.scss` files that are responsible for `grid layout` should be located under `/layouts` folder.

src/components/ProductList/index.tsx
```diff
@@ -2,14 +2,16 @@ import React from 'react'
 
 const ProductList = () => {
   return (
-      <section className="product-list">
-        <header>
+    <section className="product-list">
+      <a href="/product-detail">
+        <article>
```
LilBank: I have already commented that please make this a `4 x 1` `div` boxes. It's not a  good practice to use `<a>` tag over an `article` since (here)[https://stackoverflow.com/questions/13548225/what-is-htmls-a-a-tag-default-display-type]. You can see that `<a>` tag display is an `inline`. IMO, how about:
```
 section
    article
      header
        title
        paragraph
      a
    article
    article
    article


```

LilBank: And this should become a component called `ProductView`.

src/components/ProductView/index.tsx
LilBank: This must still be contained in a `div` since they are separated by `grid` display in `product-view`. Do you agree? @kimvcp

src/components/ProductView/product-detail.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+const ProductDetail = () => {
+  return (
+    <section className="product-detail">
```
LilBank: Let's use `div` for this first.

src/pages/product-detail.tsx
```diff
@@ -0,0 +1,19 @@
+import Footer from '../components/Footer'
+import Navbar from '../components/Navbar'
+import ProductView from '../components/ProductView'
+
+const ProductDetail = () => (
```
LilBank: For the path, would be better to use `Product` for both path and class name. For plural of product we use `ProductList` and `Product` for singular form.

src/styles/layouts/_layout-product-detail.scss
```diff
@@ -0,0 +1,45 @@
+.layout-product-detail {
```
LilBank: `layout-product-view`

LilBank (changes requested):

src/pages/product-detail.tsx
kimvcp: Don't forget to name files under `/pages` in lower case. 
```
File name -> product.tsx
Component name -> Product
```

src/components/ProductView/product-detail.tsx
Raksani: Do you mean have `section` inside another `div` or use `div` instead of `section`

src/components/ProductList/index.tsx
Raksani: 1. Could you please detail about `4 x 1 div`? 
2. about the `<a>` tag, I've check this article before [https://stackoverflow.com/questions/38367002/make-entire-article-tag-a-link](url) so that I've done like this. What if we fix this problem by adding `display: inline-block` instead?
3. The code you gave is very interesting, but it's mean we cannot click the title right? or we should back to have `button` to click instead
4. Which component? we already use `ProductView` to show only on the (old) `ProductDetail` page.

src/pages/product-detail.tsx
Raksani: @LilBank So, you mean the name of this page should be `Product`, right? which import `ProductView` component. FYI, we use the name `Products` for the page of (old) `ProductList`. This is not going to make confusion, right? 🤨

src/styles/layouts/_layout-product-detail.scss
Raksani: Which name should we use the name of `page` or `component`? Because `ProductView` is the name of `component`

src/components/ProductCard/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import Link from 'next/link'
+
+const ProductCard = () => {
+  return (
+    <div className="product-card">
+      <Link href="/product-view">
```
kimvcp, line 7: Since we are going to pass some props to each product item. I would consider use `Next router` instead

src/components/ProductCard/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+        <a className="product-card__link">
+          <div className="product-card__content">
+            <span>
+              Click Me!
+            </span>
```
kimvcp, line 12: How about one line instead?
```suggestion
            <span>Click Me!</span>
```

src/components/ProductCard/index.tsx
LilBank: This PR is actually for ProductView but the thing here is just a mockup so it could be implemented in future.

src/pages/product-view.tsx
```diff
@@ -0,0 +1,19 @@
+import Footer from '../components/Footer'
```
kimvcp: Please import `react` on the first line. We are going to use its variable in the future.  
Ref: https://github.com/vercel/next.js/issues/12964#issuecomment-629642157

src/components/ProductCard/index.tsx
kimvcp, line 7: Since we are going to pass some props to each product item. I would consider using Next Router instead.

src/components/ProductCard/index.tsx
kimvcp, line 12: How about one line instead?
```suggestion
        <span>Click Me!</span>
```

src/components/ProductCard/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import Link from 'next/link'
+
+const ProductCard = () => {
+  return (
+    <div className="product-card">
+      <Link href="/product-view">
+        <a className="product-card__link">
```
Raksani, line 8: @LilBank Regarding the design, I think we have nothing to click, apart from the `<Link>` to page `ProductView` page.

src/components/ProductCard/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+          <div className="product-card__content">
+            <span>
+              Click Me!
+            </span>
+          </div>
+        </a>
+      </Link>
+    </div>
+  )
+}
```
kimvcp: 1. Since we are going to pass some data such as id when routing to another page. I would consider using `Next Router` instead.
2. Do you think we should use this path name ('/product-view'). How about making it more unique by adding product id instead ('/products/{id}')
3. When there is only a small text, I personally like to align it in one line (`<span>Click Me!</span>`) instead. What do you think?
```suggestion
interface ProductCard {
  id: string
  // TODO: Other props will be added
}

const ProductCard = ({ id }: ProductCard) => {
  const router = useRouter()

  const handleProductCardClick = () => {
    router.push(`/products/${id}`)
  }

  return (
    <div className="product-card">
      <a className="product-card__link" onClick={handleProductCardClick}>
        <div className="product-card__content">
          <span>Click Me!</span>
        </div>
      </a>
    </div>
  )
}
```

src/components/ProductContent/ProductMedia.tsx
```diff
@@ -0,0 +1,18 @@
...
+const ProductMedia = () => {
+  return (
+    <div className="product-media">
+      <header className="product-media__header">
+        <h3 className="product-media__title">
+          Images
+        </h3>
+      </header>
+      <div className="product-media__image-container">
+        <div className="product-media__image"></div>
```
Raksani, line 12: Should we change this to ``?

src/components/ProductList/index.tsx
```diff
@@ -1,15 +1,13 @@
 import React from 'react'
+import ProductCard from '../ProductCard'
 
 const ProductList = () => {
+  const articles = new Array(20).fill(<ProductCard />)
```
kimvcp, line 5: 1. Since this is typescript, do you think we should specify type for every variables?
2. Although this is just a mocking. Shall we be more specific by adding `id` and `key` for each `ProductCard` 
```suggestion
  const articles = new Array<JSX.Element | null>(20)
    .fill(null)
    .map((productCard: any, index: number) => (
      <ProductCard key={index} id={(index + 1).toString()} />
    ))
```

Raksani, line 5: I just found that the solutions based on fill (B,C) are fast on Chrome and Safari but surprisingly slowest on Firefox for big arrays. Please check this [https://stackoverflow.com/questions/1295584/most-efficient-way-to-create-a-zero-filled-javascript-array](url)

src/styles/components/_product-media.scss
```diff
@@ -0,0 +1,32 @@
...
+    margin: 0;
+  }
+
+  &__image-container {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    height: rem(200px);
+    width: rem(200px);
```
Raksani: I think the category `display` and `box model` can be grouped together. 
Ref: [https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/]()

src/styles/components/_product-card.scss
```diff
@@ -0,0 +1,31 @@
+.product-card {
+  display: inline-block;
+  width: 50%;
+  height: rem(200px);
+  padding: rem(20px 8px);
+  box-sizing: border-box;
```
kimvcp, line 5: According to our agreed [ref](https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/), please move this `box-sizing` below your `display: inline-block`

src/styles/components/_product-card.scss
```diff
@@ -0,0 +1,31 @@
...
+    width: 33%;
+  }
+
+  @include respond-above('xl') {
+    width: 25%;
+  }
+
+  &__content {
+    height: 100%;
+    background-color: orangered;
```
kimvcp: `background-color` property should be under `border` property, isn't it?

src/styles/components/_product-detail.scss
```diff
@@ -0,0 +1,20 @@
+.product-detail {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: rem(16px);
+  box-sizing: border-box;
```
kimvcp, line 5: `box-sizing` should be above `padding`

src/styles/components/_product-information.scss
```diff
@@ -0,0 +1,20 @@
+.product-information {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: rem(16px);
+  box-sizing: border-box;
```
kimvcp, line 5: Same as this [comment](https://github.com/Cobda/cobda-web/pull/6/files#r571752547)

src/styles/components/_product-media.scss
```diff
@@ -0,0 +1,30 @@
+.product-media {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: rem(16px);
+  box-sizing: border-box;
```
kimvcp, line 5: Same as this [comment](https://github.com/Cobda/cobda-web/pull/6/files#r571752547)

src/styles/components/_product-media.scss
```diff
@@ -0,0 +1,30 @@
...
+  &__title {
+    margin: 0;
+  }
+
+  &__image-container {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: rem(200px);
+    width: rem(200px);
```
kimvcp: If we want to strictly follow this [ref](https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/), please put `width` before `height`
```suggestion
    width: rem(200px);
    height: rem(200px);
```

kimvcp (approved):

LilBank (approved):

PR #14, opened 2021-02-01
[ch108] As a user, I can see a user navigation bar

kimvcp:
https://app.clubhouse.io/cobda/story/108/as-a-user-i-can-see-a-user-navigation-bar

## Why?

This PR is for making a profile picture on the rightest of the navigation bar.

## Changes

- Navigation bar is now separated into 3 areas (Brand, Menu, Authentication).
- When user sign in, profile avatar appear instead of the authentication buttons.

## Screenshot(s)

**Before** 

Screen Shot 2564-02-01 at 15 23 31

**After**

When there is no user.
Screen Shot 2564-02-01 at 15 20 38

When there is a user.
Screen Shot 2564-02-01 at 15 21 59


Kittinske15 (approved):

Kittinske15 (approved):

src/components/Navbar/Account.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+import { useRouter } from 'next/router'
+
+interface NavbarAccount {
+  isUserExist?: boolean
```
LilBank: How about `isUserSignedIn`? IMO, **exist** is more meaningful with checking null of something.

src/components/Navbar/Account.tsx
```diff
@@ -0,0 +1,47 @@
...
+            Sign In
+          </button>
+          <button
+            className="navbar__account-button"
+            onClick={handleSignUpClick}>
+            Sign Up
+          </button>
+        </>
+      )}
+    </div>
```
LilBank: What do you think of this? Create a variable for those elements for better readability. 🤔
```suggestion
    <div className="navbar__account">
      {isUserExist ? (navbarAvatar) : (navbarAuthentication)}
    </div>
```

src/styles/components/_navbar.scss
```diff
@@ -29,4 +21,44 @@
...
+
+  // TODO: for future implementation
+  &__account-button {
+  }
+
+  &__avatar-button {
+    width: rem(48px);
+    height: rem(48px);
+    border-radius: 50%;
+  }
```
LilBank: How about a `__button` with 2 modifiers(`&--avatar`, `&--account`, maybe other names)? 🤔

src/styles/components/_navbar.scss
```diff
@@ -29,4 +21,44 @@
   @include respond-above('xl') {
     padding: 0 rem(40px);
   }
+
+  &__menu-item,
+  &__authentication-button {
```
LilBank: I don't see any usage for `authentication-button`, do you mean `&__button--account`?

src/components/Navbar/Account.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  const handleAvatarClick = () => {
+    // TODO: Expand dropdown or sth
+  }
+
+  return (
+    <div className="navbar__account">
+      {isUserExist ? (
+        <button className="navbar__avatar-button" onClick={handleAvatarClick}>
```
LilBank: I don't think avatar is a button. It is just an image displaying the user profile. How a normal `div` for this?

LilBank: But if it's going to be dropdown then I understand the usage of `button`. But then is it really `button`?

src/styles/components/_navbar.scss
```diff
@@ -29,4 +21,44 @@
...
+  // TODO: for future implementation
+  &__menu-item {
+  }
+
+  &__link {
+    text-decoration: none;
+  }
+
+  &__account {
+    display: flex;
```
LilBank: Just `display: flex` does nothing. Are you going to add any alignments or others?

LilBank (changes requested):

src/components/Navbar/Account.tsx
kimvcp: Sure, it looks a lot cleaner. But I would follow how "Airbnb" names their methods that return JSX partials. Is it okay for you?
Ref: https://github.com/airbnb/react-dates/blob/master/src/components/DateRangePicker.jsx#L317

src/styles/components/_navbar.scss
kimvcp: Oh this is typo 😅

src/components/Navbar/Logo.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+const NavbarLogo = () => {
+  return (
+    <div className="navbar__logo">
+      {/* TODO: Use logo instead*/}
```
Raksani: Does the logo appear everytime or only on the mobile site?

Raksani:
Does `Navbar` have a button to change language? I did not see one on it 👀 or we will implement it later.

src/components/Navbar/Logo.tsx
kimvcp: Per our last conversation, we need two logos in total, one for mobile and another for desktop. Thus, we will handle the logic when we have all the logos since we can simply do this
`yah`

kimvcp:
> Does `Navbar` have a button to change language? I did not see one on it 👀 or we will implement it later.

This PR is mainly for adding `Avatar`, others will be implemented later

src/components/Navbar/Account.tsx
kimvcp: Yep, it's going to be a dropdown button but since we are going to import this component from `bootstrap` or `material-ui` Can we keep only a button for now?

src/components/Navbar/Account.tsx
LilBank: According to your reference, I can only see that they don't even include conditional rendering at `render()` rather they put it on the functions instead. Furthermore, on `return()` they even put a component there instead of the components body there. In addition, from React's [docs](https://reactjs.org/docs/conditional-rendering.html), you can see at the render/return they keep it in one **LOC**. I don't think it's a good idea to use conditional parameter in your way. Does my opinion changes yours?

src/components/Navbar/Account.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+import { useRouter } from 'next/router'
+
+interface NavbarAccount {
+  isUserSignedIn?: boolean
```
LilBank, line 5: I have another better name. How about `isLoggedIn`?

src/components/Navbar/Account.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const handleSignUpClick = () => {
+    router.push('/sign-up')
+  }
+
+  const handleAvatarClick = () => {
+    // TODO: Expand dropdown or sth
+  }
+
+  const renderAvatarButton = () => (
+    <button className="navbar__button-avatar" onClick={handleAvatarClick}>
```
LilBank: This css class is not working. You mean this?
```suggestion
    <button className="navbar__button navbar__button--avatar" onClick={handleAvatarClick}>
```

src/components/Navbar/Account.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <button className="navbar__button--sign-up" onClick={handleSignUpClick}>
+        Sign Up
+      </button>
+    </>
+  )
+
+  return (
+    <div className="navbar__account">
+      {!isUserSignedIn ? renderAvatarButton() : renderAuthenticationButton()}
+    </div>
```
LilBank: What do you think about this? 🤔
```suggestion
  <div className="navbar__account">
    {isLoggedIn
      ? renderAvatarButton()
      : renderAuthenticationButton()
    }
  </div>
```
See here: https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator

src/styles/components/_navbar.scss
```diff
@@ -29,4 +21,46 @@
...
+  }
+
+  &__button {
+    &--sign-in {
+      // TODO: for future implementation
+    }
+
+    &--sign-up {
+      // TODO: for future implementation
+    }
```
LilBank, line 56: I want you to check with Github's navbar, they use's `<a>` for both. What do you think?

src/styles/components/_navbar.scss
```diff
@@ -29,4 +21,46 @@
   @include respond-above('xl') {
     padding: 0 rem(40px);
   }
+
+  &__menu-item,
+  &__button--sign-in {
```
LilBank: FYI, you can actually just use `&__button`. No need for modifier since if there's `&__button--sign-in` and `&__button--sign-up` it will know who will be the last child.

LilBank (changes requested):

src/components/Navbar/Account.tsx
kimvcp: Yeah for larger expression we do split into many lines like what you suggested, but in this case there are no `props` or other long expression. I personally prefer one line when there is only a block of text.

src/components/Navbar/Account.tsx
kimvcp: I personally name variables specifically since there will be no room for confusions. Can we go with something more specific like `isUserLoggedIn`? Can I know your opinion? @Kittinske15 @Raksani 

PS: I only use `isLoggedIn` when it is followed by a subject like `User.isLoggedIn`

kimvcp: Or else I can make an interface name `User` and use your suggestion (`isLoggedIn`)
```
interface User {
  isLoggedIn: boolean
  // other required props
  ...
} 
```

Raksani: In my opinion, `LoggedIn` relates to `Log in` button on our design. Anyway, either `isLoggedIn` or `isUserLoggedIn` looks fine to me.

src/components/Navbar/Account.tsx
kimvcp: Firstly, I agreed that we should group all conditional rendering into a function/variable. 
Secondly, what I followed from `Airbnb` is how they name their methods not how they adjust their component's style.

This is what I applied.
Screen Shot 2564-02-02 at 16 32 23

Another example inside their [DayPicker](https://github.com/airbnb/react-dates/blob/master/src/components/DayPicker.jsx)

Screen Shot 2564-02-02 at 17 56 36

Raksani:
After I looked around, I think you should choose between `SignIn` or `LogIn`. Regarding to our design has two words which are “log in” and “sign up”.

src/components/Navbar/Account.tsx
kimvcp: Thanks 😅

kimvcp (commented):
> After I looked around, I think you should choose between `SignIn` or `LogIn`. Regarding to our design has two words which are “log in” and “sign up”.

I prefer `SignIn` over `LogIn` since we have `SignOut` and `SignUp` as its relation.

src/components/Navbar/Account.tsx
LilBank: Lets keep it like this first by just renaming 👍

LilBank (approved):

Raksani (approved):

PR #15, opened 2021-02-07
[ch107] As a user, I can see Sign in Screen

Kittinske15:
https://app.clubhouse.io/cobda/story/107/as-a-user-i-can-see-sign-in-screen

## Why?

This PR is  leads to ``Sign In`` screen. And accessible by clicking ``Sign In`` button in home page.

## Changes

- Create ``Sign In`` screen.
- Add ``Sign In`` button which is link directly to ``Sign In`` screen.
- Add ``Login`` button for login user account.

## Screenshot(s)

**Before**
image

**After**
image

### DIFF
--- a/src/components/SignInForm/index.tsx
+++ b/src/components/SignInForm/index.tsx
@@ -1,12 +1,12 @@
 import React from 'react'
-import Form from '../Form'
+import Form from './Form'
 
 const SignInForm = () => {
   return (
     <section className="sign-in-form">
       <header className="sign-in-form__header">Sign In</header>
       <div className="sign-in-form__body">
-        <Form title="Personal Detail" href="/" />
+        <Form />
       </div>
       <footer className="sign-in-form__footer">
         Don't have an account? Create here
--- a/src/components/SignInForm/index.tsx
+++ b/src/components/SignInForm/index.tsx
@@ -11,14 +11,16 @@ const SignInForm = () => {
       </div>
       <footer className="sign-in-form__footer">
         <div className="sign-in-form__footer-item">
-          Don't have an account?
+          <p className="sign-in-form__description">Don't have an account?</p>
           <Link href="/sign-up">
-            <a className="sign-in-form__link">Create here</a>
+            <a className="sign-in-form__link">Create an account</a>
           </Link>
         </div>
-        <Link href="">
-          <a>Forgot you password?</a>
-        </Link>
+        <div className="sign-in-form__footer-item">
+          <Link href="">
+            <a>Forgot you password?</a>
+          </Link>
+        </div>
       </footer>
     </section>
   )
--- a/src/components/SignInForm/index.tsx
+++ b/src/components/SignInForm/index.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import Link from 'next/link'
 import Form from './Form'
 
 const SignInForm = () => {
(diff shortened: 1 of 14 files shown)


src/styles/components/_sign-in-form.scss
```diff
@@ -0,0 +1,29 @@
+@import '../abstracts/_abstracts-dir';
+
+.sign-in-form {
+  display: flex;
+  flex-direction: column;
+  background-color: $background;
```
LilBank: Please rename the variable `$background` to its original color.

src/styles/components/_sign-in-form.scss
```diff
@@ -0,0 +1,29 @@
...
+    background-color: greenyellow;
+    margin-bottom: rem(24px);
+    padding: rem(24px) 0;
+  }
+
+  &__body {
+    margin-bottom: rem(24px);
+  }
+
+  &__footer {
```
LilBank, line 20: Currently the `flex-direction` is row. How about adding `column` so **forgot password** stay below **Don't have an account?** 🧐

src/components/SignInForm/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <div className="sign-in-form__body">
+        <Form />
+      </div>
+      <footer className="sign-in-form__footer">
+        Don't have an account?
+        <Link href="/sign-up">
+          <a className="sign-in-form__link">Create here</a>
+        </Link>
+        <Link href="">
+          <a className="sign-in-form__link">Forgot password</a>
```
LilBank: What do you think of **Forgot your password?** 🤔

src/components/SignInForm/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <section className="sign-in-form">
+      <header className="sign-in-form__header">Sign In</header>
+      <div className="sign-in-form__body">
+        <Form />
+      </div>
+      <footer className="sign-in-form__footer">
+        Don't have an account?
+        <Link href="/sign-up">
+          <a className="sign-in-form__link">Create here</a>
```
Raksani: For easy understanding, I think we can simply use “Sign Up” or “Create account”, what do you think?

src/components/SignInForm/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+        </Link>
```
LilBank: This should be group together though.

LilBank (changes requested):

src/components/SignUpForm/index.tsx
```diff
@@ -1,12 +1,12 @@
...
 const SignUpForm = () => {
   return (
     <section className="sign-up-form">
       <header className="sign-up-form__header">Register</header>
       <div className="sign-up-form__body">
-        <Form title="Personal Detail" href="/sign-up-success" />
+        <Form />
       </div>
       <footer className="sign-up-form__footer">
         Already have account? Sign in
```
Raksani, line 12: I was wondering if you could make `Sign in` clickable?

src/components/SignInForm/index.tsx
Kittinske15: I think for all the sentence from ''Don't have an account? Create here'' is better than ''Don't have an account? Create account'' via redundant words in the sentence.

Raksani: Let's see these examples >> [https://uxdesign.cc/designing-a-user-friendly-login-25855ae0cc88](). And I've never seen 'Create here' before. Do you have any references to share, in case I miss something 👀 ?

Kittinske15: Good reference, so I'll change into 'create an account' 😁

PR #17, opened 2021-02-10
[ch179] Add files to connect to Docker

Raksani:
https://app.clubhouse.io/cobda/story/179/automated-build-docker-image

## Why?

To build Docker image we need Dockerfile to run the command so that every update on branch `development` will activate auto-build Docker image on Docker Hub. And processes are available to real-time monitor on Docker Hub repository.
Docker Hub: https://hub.docker.com/repository/docker/cobda/cobda

## Changes

- Our website is ready to use as **Docker Image** on **Docker Hub**
- Continuous Deployment is working successfully

## Screenshots
Config build rules
Screenshot 2564-02-10 at 20 46 11

After click `save and build`, automated builds log can be monitored via Docker Hub repository 
Screenshot 2564-02-10 at 20 46 28
Screenshot 2564-02-10 at 20 46 50
Screenshot 2564-02-10 at 20 47 58

Build Success
Screenshot 2564-02-15 at 00 02 06
Show on specific port (:8000)
Screenshot 2564-02-10 at 20 14 45


Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM node:current-alpine AS base
+WORKDIR /base
+COPY package*.json ./
+RUN npm install
```
LilBank: I was wondering if we can use `yarn` to speed up the overall process. This [link](https://medium.com/hackernoon/using-yarn-with-docker-c116ad289d56) provide instructions while this [link](https://medium.com/@iamnayr/a-multi-part-analysis-of-node-docker-image-sizes-using-yarn-vs-traditional-npm-2c20f034c08f) shows the comparison. Please kindly have a look and if possible let's use `yarn`.

.dockerignore
```diff
@@ -0,0 +1,3 @@
+.next
```
LilBank: Are we suppose to add `node_modules` too? How about `.github` and `.eslintcache`? Try turning on hidden files in your editor and see which files should be ignored.

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM node:current-alpine AS base
```
LilBank: Would be nice to name this stages 🥰 
See [here](https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages)

LilBank (changes requested):

Dockerfile
Raksani: Very interesting, I decide to use `npm` as a default so that our `dockerfile` is surely worked. Anyway, I've to modify this  a bit to support `yarn`.

Dockerfile
Raksani: From the [Docker document](https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages). "By default, the stages are not named. However, you can name your stages, by adding an AS <NAME>" So I named it as `BASE`. Anyway, it's really optional.

.dockerignore
```diff
@@ -0,0 +1,61 @@
...
+.node_repl_history
+
+# Output of 'npm pack'
+*.tgz
+
+# Yarn Integrity file
+.yarn-integrity
+
+# Git
+.git
```
LilBank: Please add a newline.

.dockerignore
```diff
@@ -0,0 +1,61 @@
...
+build/Release
+
+# Typescript v1 declaration files
+typings/
+
+# Optional npm cache directory
+.npm
+
+# Optional eslint cache
+.eslintcache
```
LilBank: Quick reminder, I think this should already be ignored by `.gitignore` so no need to put this here. Please also check for other duplicates.

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# Production image, copy all the files and run next
+FROM node:current-alpine AS runner
+ENV NODE_ENV=production
+WORKDIR /app
+
+# Copy from build image
+COPY --from=builder /build/package*.json ./
+COPY --from=builder /build/node_modules ./node_modules
+COPY --from=builder /build/.next ./.next
+COPY --from=builder /build/public ./public
```
LilBank: This should leave us 4 layers of `COPY`. According to [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers), we should reduce number of layers and chain `COPY` instead by [this](https://stackoverflow.com/questions/30256386/how-to-copy-multiple-files-in-one-layer-using-a-dockerfile).

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Expose the listening port
+EXPOSE 3000
+# Run yarn start script when container starts
+CMD ["yarn", "start"]
```
LilBank: Please add a newline.

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Set working directory
+WORKDIR /base
+
+# Copy package.json and package-lock.json before other files
+# Utilise Docker cache to save re-installing dependencies if unchanged
+COPY package*.json yarn.lock ./ 
+
+# Install dependencies
+RUN yarn install --frozen-lockfile
```
LilBank, line 13: Quick question, should we also pass `--production` too?

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Copy all files
```
LilBank: How about this?
```suggestion
# Copy installed dependencies to current working directory
```

Dockerfile
```diff
@@ -0,0 +1,39 @@
+# Base on offical Node.js Alpine image
+FROM node:current-alpine AS base
+
+# Set working directory
+WORKDIR /base
+
+# Copy package.json and package-lock.json before other files
```
LilBank:
```suggestion
# Copy package.json and yarn-lock.json to [SOME_PATH] for installation
```

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# Utilise Docker cache to save re-installing dependencies if unchanged
```
LilBank:
```suggestion
# This is to utilise Docker cache to save re-installing dependencies if unchanged
```

Dockerfile
LilBank: Quick question, since this is the part where we install dependencies, should we also add `ENV NODE_ENV=production` in this stage?

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+COPY package*.json yarn.lock ./ 
+
+# Install dependencies
+RUN yarn install --frozen-lockfile
+
+# Copy all files
+COPY . .
+
+# Build app
+FROM base AS builder
```
LilBank, line 19: Since above we add the comment to explain each step, would be nice to have it here too 🥰

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ENV NODE_ENV=production
+WORKDIR /build
+COPY . .
+COPY --from=base /base/node_modules ./node_modules
```
LilBank: If my understandings are correct, the above command is for copying everything in the `base` to `build`. Then why are you copying the `node_modules` again? 🤔

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Copy from build image
+COPY --from=builder /build/package*.json ./
+COPY --from=builder /build/node_modules ./node_modules
+COPY --from=builder /build/.next ./.next
+COPY --from=builder /build/public ./public
+
+# Expose the listening port
+EXPOSE 3000
+# Run yarn start script when container starts
```
LilBank:
```suggestion
EXPOSE 3000

# Run yarn start script when container starts
```

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+COPY . .
+COPY --from=base /base/node_modules ./node_modules
+RUN yarn build
+
+# Production image, copy all the files and run next
+FROM node:current-alpine AS runner
+ENV NODE_ENV=production
+WORKDIR /app
+
+# Copy from build image
```
LilBank: Can you please make this sentence more meaningful?
```suggestion
# Copy built files to ...
```

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# Utilise Docker cache to save re-installing dependencies if unchanged
+COPY package*.json yarn.lock ./ 
+
+# Install dependencies
+RUN yarn install --frozen-lockfile
+
+# Copy all files
+COPY . .
+
+# Build app
```
LilBank:
```suggestion
# Build image, for building the app.
```

Dockerfile
```diff
@@ -0,0 +1,39 @@
+# Base on offical Node.js Alpine image
```
LilBank:
```suggestion
# Base image, for installing the dependencies.
```

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Build app
+FROM base AS builder
+ENV NODE_ENV=production
+WORKDIR /build
+COPY . .
+COPY --from=base /base/node_modules ./node_modules
+RUN yarn build
+
+# Production image, copy all the files and run next
```
LilBank:
```suggestion
# Production image, [DESCRIPTION]
```

LilBank (changes requested):

Dockerfile
Raksani: Just `Utilize` is fine. Like `Copy`

LilBank: Utilize or Utilise?

Dockerfile
Raksani: `COPY` them individually, rather than all at once. This ensures that each step’s build successfully. It’s easy to maintain and monitor. I prefer to keep it like this. You can see more examples in file changed of this [PR](https://github.com/vercel/next.js/pull/17794)

LilBank: According to your reference, author only have 2 copies while we have 4. This is only a `COPY` command which copies the built file to the `/app` which I still don't think it should be done separately. Not like `RUN` command which some should be separated for monitoring though. What do you think? Please share some thoughts @kimvcp .

BTW, can't we just copy all the files in `builder` and just paste in `WORKDIR`?

yordis: @LilBank is a balance between running the entire stage even though you only need a section of it, vs having multi-stage where it may be the case you copy things multiple times, but it is optimized for caching the layers.

So technically, depends on which layer of invalidation you are producing.

I would say, test it if you think it is slower, this is the best I came up with https://github.com/vercel/next.js/discussions/16995#discussioncomment-132339

LilBank: Really appreciate your comment 🙇🏻‍♂️. However, I think my [comment](https://github.com/Cobda/cobda-web/pull/17/files#discussion_r575370998) is unclear so let me explain it explicitly.

My first thought was neither of each approach are slower than each other (multi `COPY`s/single `COPY`) and when I tested them the results were as expected. I have this thought because I was trying to follow **Dockerfile best practices** documentation([here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)). The document mention that we should **minimize** the number of layers in the image([see here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers)) since `ADD` and `COPY` command creates a layer. Furthermore, both approach caches will be invalid([here, 3rd bullet point](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache)) since both depends on the **built files** from the `builder` image (file changed => new commit => new build = (`cache COPY =! new COPY`) ).

So this is my thought 😅

In conclusion, both approach gives the same performance but only the generated layers are the diff. What do you think about this? Should we separate it into thin layers of one thick layer? 🤔 @yordis

PS. I'm quite new to Docker so if I miss understood anything feel free to correct me 🥰

Dockerfile
Raksani: Utilise

Dockerfile
Raksani: I put more detail like this, what do you think? 🤔 
```suggestion
# Copy files from build stage to production stage
```

Dockerfile
Raksani: I'll make it clearer, what do you think? 🤔
```suggestion
# Build stage
```

Dockerfile
Raksani: I'd make it's synchronized to what I've mentioned in earlier comments. What do you think? 🤔 
```suggestion
# Production stage
```

Dockerfile
Raksani: Yes, we should do so in every stage. I will put it in.

Dockerfile
Raksani: I'd like to explain that "we have to copy package.json and package-lock.json before other files and utilise Docker cache to save re-installing dependencies if unchanged"

Dockerfile
Raksani: Of course, I will do the same as https://github.com/Cobda/cobda-web/pull/17#discussion_r575689629

Dockerfile
LilBank: LGTM

Dockerfile
LilBank: SGTM

Dockerfile
LilBank: SGTM

Dockerfile
LilBank: How about, **we only copy package.json and yarn-lock.json for utilizing Docker cache ...**

Dockerfile
Raksani: I'll modify it to be `COPY` all files from `build stage` instead.

Dockerfile
Raksani: So, I decided to detail every stage. I'd be like this. What do you think? 🤔💭 
```suggestion
# Installation stage
# Use Node.js Alpine base image
```

Dockerfile
Raksani: About `NODE_ENV=production`, because this (build) step only relates to a production build, so we're explicitly setting the `NODE_ENV` to production.

Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Installation stage
+# Offical Node.js Alpine base image
+FROM node:lts-alpine AS base
+
+# Set working directory
+WORKDIR /base
+
+# Copy package.json and package-lock.json for utilising Docker cache
```
LilBank, line 8:
```suggestion
# Copy package.json and yarn-lock.json for utilizing Docker cache 
```

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Copy installed dependencies to current working directory
+COPY . .
+
+# Build stage
+FROM base AS builder
+ENV NODE_ENV=production
+WORKDIR /build
+COPY --from=base /base ./
+RUN yarn add --dev typescript @types/react @types/node && yarn build
```
LilBank: It seems like you've added the `devDependencies` here because of `--production` in `yarn install` command. Should we remove the `--production` so this `RUN` command only has `yarn build`?

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Install dependencies only needed
+RUN yarn install --production --frozen-lockfile
+
+# Copy installed dependencies to current working directory
+COPY . .
+
+# Build stage
+FROM base AS builder
+ENV NODE_ENV=production
```
LilBank: FYI, I've tested that currently the `ENV` isn't needed. We can remove for now.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+COPY --from=base /base ./
+RUN yarn add --dev typescript @types/react @types/node && yarn build
+
+# Production stage
+FROM node:current-alpine AS runner
+ENV NODE_ENV=production
+WORKDIR /app
+
+# Copy files from build stage to production stage
+COPY --from=builder /build/package*.json /build/node_modules /build/.next /build/public ./
```
LilBank, line 39: Quick question, couldn't we just do this? Since we're copying all the files from `/build`.
```suggestion
COPY --from=builder /build .
```

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN yarn install --production --frozen-lockfile
+
+# Copy installed dependencies to current working directory
+COPY . .
+
+# Build stage
+FROM base AS builder
+ENV NODE_ENV=production
+WORKDIR /build
+COPY --from=base /base ./
```
LilBank: A slash isn't needed since we specify only one location 😅
```suggestion
COPY --from=base /base .
```

LilBank (changes requested):

Dockerfile
Raksani: First, I used only `yarn build` command but it cause an error and failed to build image. If you would like to see the detail, you can check in our Dockerhub repository.

Dockerfile
Raksani: I've set it in case we create `docker-compose` in further development.

LilBank: If it's for future implementation then lets remove ir for now since it's unused. 😇

Dockerfile
Raksani: I intend to copy to `./` like I copied in `installation stage`

Dockerfile
Raksani: According to`Docker Document` you mentioned me earlier and my mentor’s suggestion, we can do like you suggested. But if concern about clean code and readability (so we know exactly which is used in production) we should keep it like this. I agree with both ways. So, what do you think?

Dockerfile
Raksani: Actually, I intend to copy it to `./` like in `installation stage`

.dockerignore
```diff
@@ -0,0 +1,12 @@
+Dockerfile
```
LilBank, line 2: Another comment for Docker please.

Dockerfile
```diff
@@ -0,0 +1,43 @@
+# Installation stage
+# Offical Node.js Alpine base image
+FROM node:lts-alpine AS base
+
+# Set working directory
+WORKDIR /base
+
+# Copy package.json and package-lock.json for utilising Docker cache
```
LilBank:
```suggestion
# Copy package.json and yarn-lock.json for utilising Docker cache 
```

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+FROM node:lts-alpine AS base
+
+# Set working directory
+WORKDIR /base
+
+# Copy package.json and package-lock.json for utilising Docker cache 
+# to save re-installing dependencies if unchanged
+COPY package*.json yarn.lock ./ 
+
+# Install dependencies only needed
```
LilBank:
```suggestion
# Install dependencies
```

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN yarn install --frozen-lockfile
+
+# Copy installed dependencies to current working directory
```
LilBank:
```suggestion
# Copy installed dependencies to the current working directory
```

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+COPY . .
+
+# Build stage
+FROM base AS builder
+
+# TO DO in further development
+# ENV NODE_ENV=production
+
+WORKDIR /build
+COPY --from=base /base .
```
LilBank, line 27: Since you add a comment about this on `base` and `production` stage. We should add a comment for this too.

LilBank (changes requested):

.dockerignore
Raksani: Sure.

Dockerfile
Raksani: So, do you mean this? 
```suggestion
# Copy package.json, package-lock.json, and yarn.lock for utilising Docker cache 
```

.dockerignore
Raksani: Sure.

Dockerfile
Raksani: If you mean the comment about `ENV NODE_ENV`, in this file, I add in `build` and `production` stage as in https://github.com/Cobda/cobda-web/pull/17#discussion_r575830993

Dockerfile
Raksani: So, do you mean this? 
```suggestion
# Copy package.json, package-lock.json, and yarn.lock for utilising Docker cache 
```

Dockerfile
Raksani: So, do you mean this? 
```suggestion
# Copy package.json, package-lock.json, and yarn.lock for utilising Docker cache 
```

Raksani:
According to https://github.com/Cobda/cobda-web/pull/17#discussion_r575326162, we raise these questions: 
- **Do multiple lines and a single line of the `COPY` command matter for caching in Docker?** 
- **Why don't we put every file from `build stage` to `production stage` in just one simple line?** 
In conclusion, we have 3 criteria to concern:

1. **Image layers' size**: from the experiment, the result from Dockerhub apparently says **no different** between multiple lines and a single line. 

Result: Single-line command 
Screenshot_2564-02-15_at_17 19 46

Result: Multiple-line command
Screenshot_2564-02-15_at_17 37 28

2. **Cache**: Docker looks for an existing image in its cache that it can reuse, rather than creating a new (duplicate) image. Technically, it builds from top to bottom. This means when `package.json` is changed, it will rebuild the first line `COPY --from=builder /build/package*.json ./` including the 3 following commands also which caused time consumption.
3. **Readability**: multiple-line command provide more detail than `COPY from--builder /build/ ./` in single line. So, as the cache is concerned, we decide to specify each package but still in one line.

Dockerfile
LilBank: Nope, I mean to add this `# Copy files from base stage to builder stage`

Raksani: Okay, I'll add for more clarification.

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+# Production stage
+FROM node:current-alpine AS runner
+
+# TO DO in further development
+# ENV NODE_ENV=production
+
+WORKDIR /app
+
+# Copy files from build stage to production stage
```
LilBank, line 38:
```suggestion
# Copy files from builder stage to production stage
```

LilBank (approved):

Kittinske15 (approved):

kimvcp (approved):

PR #24, opened 2021-02-13
[ch109] As a user, I can see my profile page

kimvcp:
https://app.clubhouse.io/cobda/story/109/as-a-user-i-can-see-my-profile-page

## Why?

This PR is for creating the structure of `Profile` screen. Accessible by clicking `Avatar` button in navigation bar.

## Changes

- Add layout for `Profile` screen.
- `Avatar` button is now routing to `Profile` screen.

## Screenshot(s)

**Mobile**

Screen Shot 2564-02-15 at 11 16 58

**Desktop**

Screen Shot 2564-02-15 at 11 18 01


src/components/AccountContent/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import ProductAvailable from './ProductAvailable'
+import Review from './Review'
+
+const AccountContent = () => {
+  return (
+    <section className="account-content">
+      <ProductAvailable />
+      <Review />
```
Raksani: I think it shouldn't be only a standalone `Review` 🙋🏻‍♀️

src/styles/components/_account-information.scss
```diff
@@ -0,0 +1,19 @@
+@import '../abstracts/_abstracts-dir';
+
+.account-information {
+  height: 500px;
```
Raksani:
```suggestion
  height: rem(500px);
```

src/styles/layouts/_layout-account.scss
```diff
@@ -0,0 +1,65 @@
...
+    'footer';
+
+  height: 100vh;
+
+  @include respond-above('md') {
+    grid-template-rows: rem(80px) repeat(2, 1fr) rem(250px);
+  }
+
+  @include respond-above('lg') {
+    grid-template-columns: minmax(350px, 450px) minmax(642px, 1fr);
```
Raksani: I'd be nice if you use the same unit (rem)

src/components/AccountContent/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import ProductAvailable from './ProductAvailable'
+import Review from './Review'
+
+const AccountContent = () => {
```
Raksani: Regarding the design, this section carries about products and reviews without an account or user information. So, would you mind considering another class name like AccountRelatedProduct or do you have ideas in mind?

src/components/AccountContent/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+  return (
+    <section className="account-content">
+      <ProductAvailable />
```
Raksani: FYI, we have 2 carousels:

Available products
Unavailable products (disable card just showing history)
Can you please update this section?

Ps. I think in design that's only one carousel but in the meeting, we required two.

Raksani (approved):

src/components/AccountContent/Review.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+const Review = () => {
```
LilBank: How about `AccountReview`?

src/components/AccountContent/index.tsx
LilBank: A little correction from comment above, it is unavailable but let's use the word like **Out of stock**/**Sold out**.

src/components/AccountContent/index.tsx
LilBank: +1, how about `ProductReviews` or any better ideas?

src/components/AccountContent/index.tsx
LilBank: IMO, the product and the review section **is** from the user(Account). I find it normal to use `AccountContent` though 🤔 Then should we rename the word **content** to another meaningful word?

src/styles/components/_account-information.scss
```diff
@@ -0,0 +1,19 @@
...
+
```
LilBank, line 6: Missing this here 🥸
```suggestion
  box-sizing: border-box;
```

src/styles/layouts/_layout-account.scss
```diff
@@ -0,0 +1,65 @@
...
+    grid-template-rows: rem(100px) 1fr rem(300px);
```
LilBank, line 20: Would be better to add `column-gap` here to split `sidebar` and `main` apart instead of padding.
```suggestion
    grid-template-rows: rem(100px) 1fr rem(300px);
    column-gap: some_value;
```

src/styles/layouts/_layout-account.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  @include respond-above('xl') {
+    grid-template-rows: rem(140px) 1fr rem(350px);
+  }
+
+  & > header {
+    grid-area: header;
+  }
+
+  & > aside {
```
LilBank, line 41: Currently you are setting the height at the `account-information`, how about using `padding` here and set `height: 100%` at `account-information` so it's even more responsive (imagine a user have a long/short bio)

LilBank (changes requested):

src/components/AccountContent/index.tsx
kimvcp: `AccountReview` or `ProductReview`, Which one is more comfortable for you?

src/components/AccountContent/Review.tsx
kimvcp: Since we are using the same name as mentioned [here](https://github.com/Cobda/cobda-web/pull/24#discussion_r576278004), Should we go with `ProductReview` instead?

src/components/AccountContent/index.tsx
kimvcp: Do you mean `<ProductReview />` ? Plural is generally used for arrays or collections. Component name should not be pluralized.

src/components/AccountContent/index.tsx
kimvcp: How about `ProductObsolete` for the component name?

src/components/AccountContent/index.tsx
kimvcp: Since this is the container class for products and reviews, I personally think that reviews are not part of `AccountRelatedProduct`. What if we use `AccountRelatedProduct` as a class containing `Available products` and `Out of stock products` instead?

src/styles/layouts/_layout-account.scss
```diff
@@ -0,0 +1,83 @@
...
+    @include respond-above('md') {
+      padding: 2% 6% 4%;
+    }
+
+    @include respond-above('lg') {
+      padding: rem(45px) 6% rem(75px) 0;
+    }
+
+    @include respond-above('xl') {
+      padding: rem(45px) 0 rem(75px) 0;
```
LilBank: I think we can just use `padding-right: 0;`
```suggestion
      padding-right: 0;
```

src/styles/layouts/_layout-account.scss
```diff
@@ -0,0 +1,83 @@
...
+      min-height: rem(450px);
+      padding: 4% 6% 2%;
+    }
+
+    @include respond-above('lg') {
+      padding: rem(45px) 0 rem(350px) 10%;
+    }
+    
+    @include respond-above('xl') {
+      padding: rem(45px) 0 rem(350px) 0;
```
LilBank: I think you can just do this.
```suggestion
      padding-left: 0;
```

src/styles/layouts/_layout-account.scss
```diff
@@ -0,0 +1,83 @@
...
+
+  & > header {
+    grid-area: header;
+  }
+
+  & > aside {
+    grid-area: sidebar;
+
+    // TODO: Modify these min-height, padding according to the design.
+    min-height: rem(250px);
```
LilBank: Would be better to set the `height` at `grid-template-rows`.

src/components/AccountContent/AccountProduct.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+
+const AccountProduct = () => {
+  return (
+    <div className="account-product">
+      <header className="account-product__header">
```
LilBank: Let's remove `header` for now since we don't have a `h` tag

src/components/AccountContent/AccountProduct.tsx
```diff
@@ -0,0 +1,25 @@
...
+        <ul className="account-product__list">
```
LilBank: Should we move this to a component?

src/styles/layouts/_layout-account.scss
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  & > main {
+    grid-area: main;
+
+    padding: rem(16px 16px 24px);
+
+    @include respond-above('md') {
+      padding: 2% 6% 4%;
```
LilBank: Currently you are using `padding-bottom` as `%`, how about using an exact `padding-bottom` so it is equal to `account-content > div`'s `margin-bottom`?

LilBank (changes requested):

Kittinske15 (approved):

LilBank (approved):

PR #27, opened 2021-02-19
[ch145][ch150] As a user, I can see popular search carousel and items

LilBank:
https://app.clubhouse.io/cobda/story/145/as-a-user-i-can-see-popular-search-carousel
https://app.clubhouse.io/cobda/story/150/as-a-user-i-can-see-popular-search-item

## Why?

In home page, the popular search section located under the new release item section.

## Changes

- Added carousel.
- Added carousel items.
- For now the dots will be a slider (animation can be done later).
- The height of the carousel depends on the section, so it can't be a fixed height as the design.
- The title will stick to the section since carousel paddings can't be override properly.
- Adjust `layout-section` where `main` max width will be fixed at `1400px`.
- In small screen, the buttons will be hidden.
- Drag/Slide action is enabled.
- Carousel button's only accept `String`.
- Since these items will depend on the backend, will use `div` for now.

### Before

Screen Shot 2564-02-20 at 18 47 47

### After

Desktop

Screen Shot 2564-02-20 at 18 40 16

Mobile

Screen Shot 2564-02-20 at 18 40 55


src/components/PopularItemCarousel/index.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { CarouselProvider, Slider, Slide, ButtonBack, ButtonNext, DotGroup } from 'pure-react-carousel'
+import CarouselItem from './Item'
+
+const PopularItemCarousel = () => {
```
kimvcp, line 12: How about naming it this way since we are making it reusable.

```suggestion
const Carousel = () => {
```

src/components/PopularItemCarousel/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <CarouselProvider
+      visibleSlides={3}
+      naturalSlideWidth={400}
+      naturalSlideHeight={500}
+      totalSlides={5}
+      infinite={true}
```
kimvcp: To make it reusable, do you think these `value` should be passed as `props`?

```suggestion
      visibleSlides={value}
      naturalSlideWidth={value}
      naturalSlideHeight={value}
      totalSlides={value}
      infinite={value}
```

src/components/PopularItemCarousel/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+          <CarouselItem caption={"3rd slide"} />
+        </Slide>
+        <Slide index={4}>
+          <CarouselItem caption={"4th slide"} />
+        </Slide>
+        <Slide index={5}>
+          <CarouselItem caption={"5th slide"} />
+        </Slide>
+      </Slider>
+      <ButtonBack className="buttonBack">{'<'}</ButtonBack>
```
kimvcp, line 37: According to the design below, do you think we should hide this back button (`<`) when it is showing the first slide?  

Screen Shot 2564-02-20 at 23 16 37

src/components/PopularItemCarousel/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <ButtonNext className="buttonNext">{'>'}</ButtonNext>
+      <DotGroup className="dotGroup" />
```
kimvcp, line 39: Since you are making a reusable component, can you make this `DotGroup` appear or hide according to the `props` passing from parent?
PS: this is the carousel from `AboutUs` screen where there is no `DotGroup`.
Screen Shot 2564-02-20 at 23 20 37

src/components/PopularItemCarousel/index.tsx
LilBank: For now, we'll keep it this way since we don't know how `DRY` could it be but will keep this in mind.

src/styles/components/_carousel-item.scss
```diff
@@ -0,0 +1,53 @@
...
+  align-items: center;
+  height: 100%;
+
+  &__image-container {
+    height: 100%;
+    width: 100%;
+    margin-bottom: rem(12px);
+    padding: rem(5px);
+    box-sizing: border-box;
+    box-shadow: rem(1px 1px 2px 4px) $popular-item-shadow;
```
Kittinske15: It would be better if you spacing and ordering between your CSS properties?

Kittinske15 (changes requested):

src/components/PopularItemCarousel/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const PopularItemCarousel = () => {
+  return (
+    <CarouselProvider
+      visibleSlides={3}
+      naturalSlideWidth={400}
+      naturalSlideHeight={500}
+      totalSlides={5}
+      infinite={true}
+    >
```
kimvcp:
```suggestion
    <CarouselProvider
      visibleSlides={3}
      naturalSlideWidth={400}
      naturalSlideHeight={500}
      totalSlides={5}
      infinite={true}>
```

kimvcp (approved):

src/styles/components/_carousel.scss
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  &__dot {
+    border: none;
+    background-color: $athens-gray;
+    padding: rem(2px 10px);
+
+    &:disabled {
+      background-color: #638cff;
```
Kittinske15: Should be color naming in abstract variables

Kittinske15 (changes requested):

Kittinske15 (approved):

PR #36, opened 2021-02-24
[ch155] As a user, I can see footer section

kimvcp:
https://app.clubhouse.io/cobda/story/155/as-a-user-i-can-see-footer-section

## Why?

This **PR** is for adding footer section.

## Changes

- Add a footer section containing `Logo`, `Language dropdown`, and other footer items. 
- Create another type of `Language dropdown`
- `Language dropdown` is not displayed when screen size is smaller than `md`
- Content supports both `TH` and `EN`.

## Screenshot(s)

**Before**

Screen Shot 2564-02-24 at 21 00 43

**After**

Mobile

Screen Shot 2564-02-24 at 20 59 16

Desktop

Screen Shot 2564-02-24 at 20 58 41


locales/th/common.json
```diff
@@ -1,4 +1,11 @@
 {
   "signIn": "เข้าสู่ระบบ",
-  "signUp": "สมัครสมาชิก"
+  "signUp": "สมัครสมาชิก",
+  "helpCenter": "ศูนย์ให้บริการความช่วยเหลือ",
+  "termsOfService": "เงื่อนไขการให้บริการ",
+  "privacyPolicy": "นโยบายความเป็นส่วนตัว",
+  "emergencyLine": "สายฉุกเฉิน",
```
Raksani: In the design, it uses “Emergency line” which actually means for local emergency line. For example, in [Thailand](https://en.m.wikipedia.org/wiki/Telephone_numbers_in_Thailand)

Would it be better to change it to be “Helpline”? since I saw people use the keyword, “Facebook helpline” to search for their telephone number. 

Or move it to`contact us: telephone` instead?

src/components/LanguageDropdown/index.tsx
```diff
@@ -2,18 +2,26 @@ import React, { useEffect, useRef, useState } from 'react'
...
 }
 
-const NavbarLanguage = () => {
+interface LanguageDropdown {
+  readonly componentType: ComponentType
+}
+
+const LanguageDropdown = ({ componentType }: LanguageDropdown) => {
+  const router = useRouter()
+  const { locales, locale } = router
```
Raksani, line 18: Would you mind if I ask, is `locales` a plural form or a different one?

src/styles/layouts/_layout-about-us.scss
```diff
@@ -45,5 +45,9 @@
 
   & > footer {
     grid-area: footer;
+
+    display: flex;
+    justify-content: center;
+    box-shadow: 0 0 rem(20px) 0 box-shadow-opacity(0.16);
```
Raksani, line 51: Can we use `@mixin` to make it reusable? since I saw you applied this pattern to many classes.

src/components/Footer/index.tsx
```diff
@@ -1,15 +1,62 @@
...
+  Footer
+}
+
+const Footer = () => {
+  const { t } = useTranslation('common')
+
+  const renderLogoColumn = () => (
+    <div className="footer__logo-container">
+      <Logo />
+      <LanguageDropdown componentType={ComponentType.Footer} />
```
LilBank: `componentType` sounds strange; For example: `LanguageDropdown` has `componentType` `Footer`. What do you think about `childOf` or `parent` or any other meaningful names?

src/components/Footer/index.tsx
```diff
@@ -1,15 +1,62 @@
 import React from 'react'
 import Link from 'next/link'
+import Image from 'next/image'
+import Logo from '../Navbar/Logo'
```
LilBank: How about move `Logo` out as an element instead of staying under `Navbar`?

src/components/Footer/item.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import Link from 'next/link'
+
+interface FooterItem {
+  readonly label: string
+  readonly href?: string
+  readonly isTitleText?: boolean
+  readonly isEmergencyText?: boolean
```
LilBank: From `enum ComponentType`, we should create for these too. Like, `ContextType` or any other meaningful names?

src/components/Footer/item.tsx
```diff
@@ -0,0 +1,33 @@
...
+  
+  const linkClassName: string = isEmergencyText
+    ? 'footer__link footer__link--danger'
+    : 'footer__link'
+
+  const renderListItem = () =>
+    isTitleText ? (
+      <h3 className="footer__list-title">{label}</h3>
+    ) : (
+      <Link href={href || '#'}>
```
LilBank: Quick question, what does the hashtag refer to?

src/components/Footer/item.tsx
```diff
@@ -0,0 +1,33 @@
...
+interface FooterItem {
+  readonly label: string
+  readonly href?: string
+  readonly isTitleText?: boolean
+  readonly isEmergencyText?: boolean
+}
+
+const FooterItem = ({ label, href, isTitleText, isEmergencyText }: FooterItem) => {
+  
+  const linkClassName: string = isEmergencyText
```
LilBank: This could also be move into a function since it's not reusable.

src/components/LanguageDropdown/index.tsx
```diff
@@ -2,18 +2,26 @@ import React, { useEffect, useRef, useState } from 'react'
 import { useRouter } from 'next/router'
 import Image from 'next/image'
 import LanguageItem from '../LanguageItem'
+import { ComponentType } from '../Footer'
 
 export enum LocaleCode {
   English = 'EN',
-  Thai = 'TH'
+  Thai = 'TH',
```
LilBank: Intentional?
```suggestion
  Thai = 'TH'
```

src/components/LanguageDropdown/index.tsx
```diff
@@ -26,7 +34,9 @@ const NavbarLanguage = () => {
   const handleDropdownToggle = () => setMenuOpen((prevState) => !prevState)
 
   const handleMouseClick = (event: MouseEvent) => {
-    const isFocused: boolean | undefined = flagRef.current?.contains(event.target as Node)
+    const isFocused: boolean | undefined = flagRef.current?.contains(
+      event.target as Node
+    )
```
LilBank: Could be in oneline I guess? Even if it's the formatter's work.

src/components/LanguageDropdown/index.tsx
```diff
@@ -46,15 +56,35 @@ const NavbarLanguage = () => {
...
+        src="/icons/united-states.svg"
+        height={imageSize}
+        width={imageSize}
+      />
     )
   }
 
+  const renderSelectedLocale = () => (
+    <>
+      <Image src="/icons/world.svg" height={16} width={16} />
```
LilBank, line 60: Since the above `Image` tag is split into multiple lines, which are we sticking with? The multiple lines one?

src/components/Navbar/Tool.tsx
```diff
@@ -1,12 +1,13 @@
 import React from 'react'
 import NavbarAccount from './Account'
-import NavbarLanguage from './Language'
+import LanguageDropdown from '../LanguageDropdown'
+import { ComponentType } from '../Footer'
```
LilBank: Should this enum become a global enum? There any many usage of this.

src/styles/components/_dropdown.scss
```diff
@@ -1,52 +1,78 @@
...
-    box-shadow: 1px 1px rem(4px) $dropdown-list-shadow;
-    width: max-content;
-    padding: rem(2px) 0;
-    border: 1px $shuttle-gray;
-    border-radius: rem(6px);
-    margin: rem(16px) 0 0;
-    background-color: $light;
-    cursor: pointer;
-  }
+.dropdown-container {
```
LilBank: Filename should match the css class. How about making the `dropdown-container` as an element within `dropdown`?

LilBank (changes requested):

locales/th/common.json
kimvcp: Since we don't have any contact number yet, I think "Helpline" is a good fit for now.

src/components/LanguageDropdown/index.tsx
kimvcp: `locales` in this case is the collection of all locales we have, but `locale` (without `s`) is the current one applying to our website. 
Further reading [here](https://nextjs.org/docs/api-reference/next/router)

This is the output:

Screen Shot 2564-02-26 at 00 49 25

src/styles/layouts/_layout-about-us.scss
kimvcp: Personally, I think we don't need to make the @mixin for this since there are many different `box-shadow` properties in our website.

src/components/LanguageDropdown/index.tsx
kimvcp: Single line looks better.

src/components/Footer/item.tsx
kimvcp: At first I decided to use this `#` as the default one but seems like keeping it empty is a better idea.

src/components/Footer/index.tsx
kimvcp: I'm good to go with `parent`.

src/components/Footer/item.tsx
kimvcp: If we make it as a function, then we would have to store its value into variable and invoke it for returning string right? 
If so, what if we just store a string value into a variable and use it without invoking any function? 

Example code from `development` branch
Screen Shot 2564-02-26 at 03 47 44

Kittinske15 (approved):

src/styles/components/_footer.scss
```diff
@@ -1,32 +1,102 @@
...
+          display: block;
+          margin-top: rem(16px);
+        }
+
+        @include respond-above('lg') {
+          margin-top: rem(32px);
+        }
+      }
+    }
+  }
```
LilBank:
```suggestion
    .dropdown {
      display: none;

      @include respond-above('md') {
        display: block;
        margin-top: rem(16px);
      }

      @include respond-above('lg') {
        margin-top: rem(32px);
      }
    }
```

src/styles/components/_footer.scss
```diff
@@ -1,32 +1,102 @@
...
+
+  &__email-container {
+    display: flex;
+    align-items: flex-end;
+
+    & > * {
+      &:last-child {
+        margin-left: rem(8px);
+      }
+    }
```
LilBank:
```suggestion
    .footer__list-item {
      margin-left: rem(8px);
    }
```

src/styles/components/_dropdown.scss
```diff
@@ -48,5 +62,17 @@
...
+      font-weight: bold;
+    }
+
+    @include respond-above('md') {
+      font-size: rem(12px);
+    }
+
+    @include respond-above('lg') {
+      font-size: rem(16px);
+    }
```
LilBank:
```suggestion
    @include respond-above('md') {
      font-size: rem(12px);
    }

    @include respond-above('lg') {
      font-size: rem(16px);
    }

    &--bold {
      font-weight: bold;
    }
```

LilBank (changes requested):

LilBank (approved):

PR #39, opened 2021-02-26
[ch133][ch137] As a user, I can see carousel and items of Quote box in About us screen

kimvcp:
https://app.clubhouse.io/cobda/story/133/as-a-user-i-can-see-quote-box-in-about-us-screen
https://app.clubhouse.io/cobda/story/137/as-a-user-i-can-see-quote-section-displaying-carousel-in-about-us-screen

## Why?

This **PR** is for adding Quote section under Goal section.

## Changes

- Add quote section containing `Carousel` and `Carousel Item`.
- `Carousel` is displaying 1 slide at a time.
- `Carousel Item` contains profile and quote box.
- Drag/Slide action is enabled.
- Indicators and buttons (`<`, `>`) are hidden for small screen, but auto playing is enabled instead.
- Locale supports both `EN` and `TH`.

## Screenshot(s)

**Before**

Screen Shot 2564-02-28 at 01 26 16
 
**After**

Mobile

Screen Shot 2564-02-28 at 03 40 54

Desktop

Screen Shot 2564-02-28 at 03 41 05


src/styles/components/_carousel.scss
```diff
@@ -58,7 +82,7 @@
     border: none;
     background-color: $mine-shaft;
     color: $light;
```
Raksani, line 94: Just a question. Should we set these buttons' outline as `none`? since the default is orange which is not going well with our signature color set 🎨
Screenshot 2564-02-28 at 17 46 17

src/styles/components/_carousel-item.scss
```diff
@@ -6,6 +6,23 @@
...
   height: 100%;
 
+  &--quote {
+    height: fit-content;
+    padding: rem(8px) 0 0 ;
+    box-shadow: 0 0 rem(20px) 0 box-shadow-opacity(0.16);
+
+    @include respond-above('sm') {
+      padding: rem(16px) 0;
+    }
```
Raksani, line 16: I think `padding:0` looks good as well and not too far from the `title`. What do you think? 🤔 
Screenshot 2564-02-28 at 17 52 19

src/styles/components/_quote-box.scss
```diff
@@ -0,0 +1,99 @@
...
+      @include respond-above('md') {
+        background-size: rem(26px 24px), rem(26px 24px);
+      }
+
+      @include respond-above('lg') {
+        background-size: rem(30px 28px), rem(30px 28px);
+      }
+
+      @include respond-above('xl') {
+        background-size: rem(36px 32px), rem(36px 32px);
```
Raksani: Actually, I tried `background-size: rem(x y)` and it applied for both components. 
```suggestion
        size: rem(14px 12px);
        position: top left, bottom right;
      }

      @include respond-above('sm') {
        background-size: rem(22px 20px);
      }

      @include respond-above('md') {
        background-size: rem(26px 24px);
      }

      @include respond-above('lg') {
        background-size: rem(30px 28px);
      }

      @include respond-above('xl') {
        background-size: rem(36px 32px);
```

Raksani (changes requested):

src/styles/components/_quote-box.scss
kimvcp: Good catch!

src/styles/components/_carousel.scss
kimvcp: For your concern, how about using `outline: thin dotted;` instead. The main purpose is for [accessibility](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/What_is_accessibility)

Further reading [here](https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/)

src/styles/components/_carousel-item.scss
kimvcp: The reason of my `padding` is to make sure that there is some space when `shadow` appeared. However, `shadow` seem to be hidden by a `Carousel`. It will surely be fixed so could I keep it there for now?

**Expected output**
Screen Shot 2564-03-01 at 01 38 42

src/components/ProfileCard/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className="profile-card profile-card--background-image">
+      <div className="profile-card__body">
+        <img src={imagePath} alt={alternate} className="profile-card__image" />
+        <div className="profile-card__label profile-card__label--name">
+          {name}
+        </div>
+        <div className="profile-card__label profile-card__label--role">
+          {role}
+        </div>
```
LilBank: Try using `figure` and `figcaption` in this. See [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption)

src/components/QuoteBox/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { t } = useTranslation('about-us')
+
+  return (
+    <div className="quote-box">
+      <header className="quote-box__header">
+        <h3 className="quote-box__title">{t('quoteSectionHeader')}</h3>
+      </header>
+      <div className="quote-box__container quote-box__container--background-image">
+        <p className="quote-box__content">{quote}</p>
+      </div>
```
LilBank: `quote-box__container` inside `quote-box` just doesn't make sense when both of them are a block element.

src/styles/components/_about-us-section.scss
```diff
@@ -1,6 +1,12 @@
+@import '../abstracts/_abstracts-dir';
+
 .about-us-section {
+  padding: rem(24px) 0;
   box-sizing: border-box;
-  padding: rem(16px) 0;
+
+  @include respond-above('sm') {
+    padding: rem(32px) 0;
+  }
```
LilBank: Screen size within `sm` - `md` shouldn't have bigger `paddings` than `< sm`. Please use the same values from `home-section`.

src/styles/components/_profile-card.scss
```diff
@@ -0,0 +1,109 @@
...
+  display: flex;
+  justify-content: flex-end;
+
+  @include respond-above('md') {
+    width: 30%;
+  }
+
+  &--background-image {
+    background: {
+      image: url('/images/blue-ellipse.png');
```
LilBank: Please rename the file to something like `profile-card-background` or any other meaningful names.

src/styles/components/_profile-card.scss
```diff
@@ -0,0 +1,109 @@
...
+  &__label {
+    &:not(:last-child) {
+      margin-bottom: rem(4px);
+
+      @include respond-above('md') {
+        margin-bottom: rem(8px);
+      }
+    }
+
+    &--name {
```
LilBank: How about renaming this modifier to `large` and `small` for the other one?

src/styles/components/_profile-card.scss
```diff
@@ -0,0 +1,109 @@
...
+      size: contain;
+      position: center;
+    }
+  }
+
+  &__body {
+    flex-direction: column;
+    box-sizing: border-box;
+    box-shadow: 0 0 rem(20px) 0 box-shadow-opacity(0.16);
+    width: fit-content;
```
LilBank: Don't see any effect of this property, please recheck.

src/styles/components/_quote-box.scss
```diff
@@ -0,0 +1,99 @@
+@import '../abstracts/_abstracts-dir';
+
+.quote-box {
+  width: 100%;
+  height: fit-content;
```
LilBank: Don't see effect of this property, please recheck.

src/styles/components/_quote-box.scss
```diff
@@ -0,0 +1,99 @@
...
+
+  @include respond-above('md') {
+    margin-top: rem(24px);
```
LilBank: How about using `margin-top` instead?

LilBank (changes requested):

src/styles/components/_profile-card.scss
kimvcp: Since we only have 2 modifiers for this label. How about `primary` and `secondary` ?

src/styles/components/_quote-box.scss
kimvcp: I think you mean `padding-top` instead

Kittinske15 (approved):

src/styles/components/_about-us-section.scss
```diff
@@ -13,4 +15,49 @@
...
+    }
+  }
+
+  &__description {
+    margin: 0;
+  }
+
+  &--light {
+    height: rem(200px);
+    background-color: teal;
```
Raksani: Please add `teal` color as a variable. And according to [the way to order CSS properties](https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/), we have to separate `color` ☺️

src/styles/components/_carousel.scss
```diff
@@ -5,35 +5,63 @@
...
+      max-width: 87vw;
+      margin: auto;
+
+      @include respond-above('sm') {
+        min-width: auto;
+        max-width: 93vw;
+      }
+
+      @include respond-above('lg') {
+        max-width: 93.5vw;
```
Raksani, line 43: I'm just curious, why we use `vw` unit?

Raksani (changes requested):

src/styles/components/_carousel.scss
kimvcp: The carousel's width from lib `pure-react-carousel` does not shrink along with the screen. `vw` is set to ensure that it will shrink to the screen properly.
PS. You can try this yourself, setting `max-width` to something larger and `min-width` accordingly

src/components/ProfileCard/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  readonly alternate: string
+  readonly name: string
+  readonly role: string
+}
+
+const ProfileCard = ({ imagePath, alternate, name, role }: ProfileCard) => {
+  return (
+    <div className="profile-card profile-card--background-image">
+      <div className="profile-card__body">
+        <img src={imagePath} alt={alternate} className="profile-card__image" />
```
LilBank: Currently the `margin` is applied on the `img`, would be better to have an `image-container`.

src/components/QuoteBoxCarousel/Item.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface CarouselItem {
+  readonly imagePath: string
+  readonly alternate: string
+  readonly name: string
+  readonly role: string
+  readonly quote: string
+}
+
+const CarouselItem = ({ imagePath, alternate, name, role, quote }: CarouselItem) => {
```
LilBank: Currently you are using arguments from this function to create `ProfileCard` and `QuoteBox`, what do you think of changing `CarouselItem` parameters to accept `ProfileCard` and `QuoteBox` instead?

src/components/QuoteBoxCarousel/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+const QuoteBoxCarousel = () => {
+  const { t } = useTranslation('about-us')
+
+  // TODO: Remove this sample and use data from backend instead
+  const carouselItemList: Array<QuoteBoxItem> = new Array(4).fill({
+    imagePath: '/images/yeezy-380.jpg',
+    alternate: 'Adidas fashion sportswear',
+    name: t('name'),
+    role: t('role'),
+    quote: t('quoteSectionContent'),
```
LilBank:
```suggestion
    quote: t('quoteSectionContent')
```

src/styles/components/_about-us-section.scss
```diff
@@ -13,4 +15,51 @@
...
+    }
+  }
+
+  &__description {
+    margin: 0;
+  }
+
+  &__light {
+    background-color: $light;
+  }
```
LilBank: Please recheck the usage of this class.

LilBank (changes requested):

src/components/QuoteBoxCarousel/Item.tsx
kimvcp: Very interesting 👏

src/components/ProfileCard/index.tsx
kimvcp: Sure, I would go with `image-wrapper` instead since `container` means to have many elements inside. In this case, we only have one. What do you think?

Further reading [here](https://css-tricks.com/best-way-implement-wrapper-css/)

LilBank (approved):

Raksani (approved):

PR #42, opened 2021-03-10
[ch129] Models CRUDS

LilBank:
https://app.clubhouse.io/cobda/story/219/models-cruds

## Why?

Add basic models to the application so as providing their endpoints.

## Changes

- Add `User`, `Product`, `Role` models
- Add `User` and `Product` endpoints

## Screenshot(s)

Please see changes.


src/entity/entities.tsx
```diff
@@ -0,0 +1,24 @@
...
+                    histories: Object[]) => ({
+  username,
+  password,
+  firstname,
+  lastname,
+  phoneNumber,
+  role,
+  products,
+  histories
+})
```
kimvcp: Just curious cause I'm not quite familiar with this kind of format, Do you think my suggestion is easier to read or should we keep it this way? BTW, both of them are fine with me.

```suggestion
export const user = (
  username: string,
  password: string,
  firstname: string,
  lastname: string,
  phoneNumber: string,
  role: Role,
  products: Object[],
  histories: Object[]
) => ({
  username,
  password,
  firstname,
  lastname,
  phoneNumber,
  role,
  products,
  histories
})
```

src/pages/api/products/[id].tsx
```diff
@@ -0,0 +1,31 @@
...
+  switch (method) {
+    case 'PATCH':
+      return res.json(handlePatch())
+    case 'DELETE':
+      return res.json(handleDelete())
+    default:
+      res.setHeader('Allow', ['PATCH', 'DELETE'])
+      res.status(405).end(`Method ${method} Not Allowed`)
+  }
+}
```
Comment (kimvcp) [inline src/pages/api/products/[id].tsx]: I have 2 concerns
1. I don't think we need to declare a function just for returning an object. If you are planing to implement it later, then  please add `TODO` for others to understand.
2. If you agree with my previous concern, please specify its type also. 
Ex. `Message` in my suggestion is the interface name

PS: My variable names are just example, you can change them to whatever you want.

```suggestion
  const patchMessage: Message = {
    message: `This is a PATCH request, it should patch id: ${id}. Currently in progress...`,
    id: id,
  }

  const deleteMessage: Message = {
    message: `This is a DELETE request, it should DELETE id: ${id}. Currently in progress...`,
    id: id,
  }

  switch (method) {
    case 'PATCH':
      return res.json(patchMessage)
    case 'DELETE':
      return res.json(deleteMessage)
    default:
      res.setHeader('Allow', ['PATCH', 'DELETE'])
      res.status(405).end(`Method ${method} Not Allowed`)
  }
}
```

src/pages/api/users/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+
+  const mockUsers = () => {
+    const userNames = ['mocking-bird', 'wood-pecker', 'black-crow']
+    const products = [product('shoe1', 1), product('shoe2', 1), product('shoe3', 1)]
+    const histories = products.reverse()
+    const users = userNames.map(name => {
+      const password = [...name].reverse().join('')
+      const splitName = name.split('-')
```
kimvcp: Please specify type for these variables. 🤓

kimvcp (changes requested):

src/pages/api/products/[id].tsx
Comment (LilBank) [reply src/pages/api/products/[id].tsx]: Added `TODO`.

src/pages/api/products/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  const handlePatch = () => {
+    return {
+      message: 'This is a PATCH request, currently in progress...',
+      context: body
+    }
+  }
+
+  // TODO: Handle the function properly
+  const handlePost = () => {
+    const products: Array<ReturnType<typeof product>> = [product('shoe1', 1), product('shoe2', 1), product('shoe3', 1)]
```
kimvcp, line 17: 🧐
```suggestion
    const products: Array<ReturnType<typeof product>> = [
      product('shoe1', 1),
      product('shoe2', 1),
      product('shoe3', 1),
    ]
```

src/pages/api/products/[id].ts
```diff
@@ -0,0 +1,33 @@
...
+  const handleDelete = () => {
+    return {
+      message: `This is a DELETE request, it should DELETE id: ${id}. Currently in progress...`,
+      id: id
+    }
+  }
+
+  switch (method) {
+    case 'PATCH':
+      return res.json(handlePatch())
```
Comment (kimvcp) [inline src/pages/api/products/[id].ts:24]: I am not quite familiar with this approach ( `res.json(handle...)` ), Do people normally use the word `handle` to name their function that **returns object**?. If so please provide me ref(s). 
Personally, It think it looks cleaner if we could possibly make something like below. What do you think?

```
  switch (method) {
    case 'POST': 
      handlePost()
      break;
    case 'PATCH': 
      handlePatch()
      break;
    default: ...
```

kimvcp (changes requested):

src/pages/api/users/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  const handlePost = () => {
+    return {
+      users: mockUsers(),
+      context: body
+    }
+  }
+
+  const mockUsers = () => {
+    const userNames: string[] = ['mocking-bird', 'wood-pecker', 'black-crow']
+    const products: Array<ReturnType<typeof product>> = [product('shoe1', 1), product('shoe2', 1), product('shoe3', 1)]
```
kimvcp, line 26: Same [here](https://github.com/Cobda/cobda-web/pull/42/files#r593735478)

src/pages/api/users/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    const histories: Array<ReturnType<typeof product>> = products.reverse()
+    const users: Array<ReturnType<typeof user>> = userNames.map(name => {
+      const password: string = [...name].reverse().join('')
+      const splitName: string[] = name.split('-')
+
+      return user(name, password, splitName[0], splitName[1], '0', Role.User, products, histories)
```
kimvcp, line 32: Do you think this is easier to read?
```suggestion
      return user(
        name,
        password,
        splitName[0],
        splitName[1],
        '0',
        Role.User,
        products,
        histories
      )
```

LilBank: No, the amount of letters on the line is not even above 120 letters.

src/pages/api/users/index.ts
LilBank: https://github.com/Cobda/cobda-web/pull/42#discussion_r593739603

src/pages/api/products/index.ts
LilBank: https://github.com/Cobda/cobda-web/pull/42#discussion_r593739603

src/pages/api/products/[id].ts
Comment (LilBank) [reply src/pages/api/products/[id].ts:24]: Other developers preferences: https://stackoverflow.com/questions/6114210/is-returning-out-of-a-switch-statement-considered-a-better-practice-than-using-b
Facts (See last paragraph, last sentence): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#description

kimvcp (approved):

Kittinske15 (approved):

PR #44, opened 2021-03-10
[ch204] As a user, I can see reCAPTCHA checkbox in Sign-up screen

Raksani:
https://app.clubhouse.io/cobda/story/204/as-a-user-i-can-see-recaptcha-checkbox-in-sign-up-screen

## Why?

To keep malicious software from engaging in abusive activities on our website, reCAPTCHA uses an advanced risk analysis engine and adaptive challenges. As a result, fake users will be blocked.

## Changes

The reCAPTCHA checkbox is located in the registration form and above the `REGISTER` button.

## Screenshot(s)

**Before**
Screenshot 2564-03-10 at 23 43 37

**After**
Screenshot 2564-03-11 at 16 27 34


src/styles/components/_form.scss
```diff
@@ -52,6 +52,12 @@
     padding: rem(4px 12px);
   }
 
+  &__recaptcha {
+    display: flex;
+    justify-content: center;
+    margin-bottom: rem(24px);
+  }
```
LilBank: You can combine `margin-bottom` with the above `&:not(:last-child)` and for the `display` properties combine them with `&__actionable`.

src/components/SignUpForm/Form.tsx
```diff
@@ -38,6 +39,9 @@ const Form = () => {
         <label className="form__input-label">Confirm Password</label>
         <input type="text" className="form__input" />
       </div>
+      <div className="form__recaptcha">
+        <ReCAPTCHA sitekey={process.env.NEXT_PUBLIC_RECAPTCHA_SITE_KEY} />
```
LilBank: What exactly is this key for?

.gitignore
```diff
@@ -69,7 +69,7 @@ typings/
 .yarn-integrity
 
 # dotenv environment variables file
-.env
+.env*
```
LilBank: Do we still need `.env.test` if you use the asterisk sign?

LilBank (changes requested):

src/components/SignUpForm/Form.tsx
Raksani: The site key is used to invoke reCAPTCHA service on your site or mobile application.

.gitignore
Raksani: Yes, I just try not to push `.env` file by accident. I'll modify it back to the original.

.env.local
```diff
@@ -0,0 +1,2 @@
+NEXT_PUBLIC_RECAPTCHA_SITE_KEY=""
+RECAPTCHA_SECRET_KEY=""
```
LilBank: Please rename it to be more meaningful.
```
GOOGLE_RECAPTCHA_SITE_KEY
GOOGLE_RECAPTCHA_SECRET_KEY
```

LilBank: Also don't forget the newline.

LilBank (changes requested):

.env.local
Raksani: For the name of `NEXT_PUBLIC`, I tend to make the token visible to the client. You can check about the exposing env variable [here](https://nextjs.org/docs/basic-features/environment-variables). Plus, the **reCaptcha** is specific enough because it's a product name of Google.

kimvcp (approved):

src/components/SignUpForm/Form.tsx
```diff
@@ -38,6 +39,9 @@ const Form = () => {
         <label className="form__input-label">Confirm Password</label>
         <input type="text" className="form__input" />
       </div>
+      <div className="form__recaptcha">
+        <ReCAPTCHA sitekey={process.env.NEXT_PUBLIC_RECAPTCHA_SITE_KEY!} />
```
LilBank, line 43: Quick question why do we need the `!`?

.env.local
```diff
@@ -0,0 +1,2 @@
+NEXT_PUBLIC_RECAPTCHA_SITE_KEY=""
```
LilBank: I don't think we need the double quote since when the value is imported to the application, it will be treat as string 🤔

Raksani: For me, Do/Don't have gives the same result. but I just put the double quote in case there's provided with bare text like `fjkldol, ef893mfkf`so that you just copy and paste it inside the double quote directly. And yes, it's a string.

.env.local
LilBank: Who's the client? We are the client though !? Then it's okay not to add the `google` prefix but better to remove the `NEXT_PUBLIC`. Try to name these variables by their sources. BTW, don't see any usage of the secret key though.

src/components/SignUpForm/Form.tsx
Raksani: To solve the problem about `undefined` type. We can now use the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). It tells TypeScript that even though something looks like it could be null, it can trust us that it's not.

kimvcp: Alternatively, we can choose to render this `<ReCAPTCHA />` component only if the `process.env.NEX...` is not `undefined` or `null`. The solution can be found in [react conditional rendering](https://reactjs.org/docs/conditional-rendering.html)

.env.local
LilBank: Then please remove, no need for extra.

.env.local
Raksani: 1. The `secret key` is definitely used in further development. Would it be nice to declare both secret and site key, wouldn't it?
2. According to the ref I provided, by default all environment variables loaded through .env.local are only available in the Node.js environment, meaning they won't be exposed to the browser. In order to expose a variable to the browser, it needs a `NEXT_PUBLIC` prefix and the site key is the public key. If you don't mind, please provide any information to support that we do not need this prefix.

src/components/SignUpForm/Form.tsx
Raksani: As I said earlier, we definitely declare the environment variable (sitekey) but in Github we don't have one, so you have to use your local env instead. Anyway, in Docker, I already declared it. Surely, the site key never be undefined.

LilBank (approved):

Kittinske15 (approved):

PR #45, opened 2021-03-11
[ch207] As a user, I can see profile image in Sign-up screen

kimvcp:
https://app.clubhouse.io/cobda/story/207/as-a-user-i-can-see-profile-image-in-sign-up-screen

## Why?

The **PR** is for adding profile image inside Sign-up form.

## Changes

- Add default profile upload button with a single file selection.
- When the image is selected, it displays to the screen. 
- Confirmation alert popup when refreshing page and browsing away.

### Before

Screen Shot 2564-03-11 at 19 15 49

### After

Screen Shot 2564-03-11 at 19 15 25

Image is selected

Screen Shot 2564-03-11 at 19 19 32


src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+      const imageUrl: string = URL.createObjectURL(files[0])
+      setSelectedImageUrl(imageUrl)
+    }
+  }
+
+  const renderProfileImage = () => {
+    const defaultProfileImage: ReactNode = (
+      <img
+        className="profile-upload__image"
+        src="/images/default-profile-image.png"
```
LilBank, line 52: People may not upload their profile image on first registration, so this name may got confuse with the actual **default profile image** in user profile. How about adding `signup` word to this name?

src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+        alt="Profile Upload Image"
```
LilBank: This should be more meaningful incase the website doesn't load this image. What do you think of **Signup profile upload image icon** or a better name?

src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+      />
+    )
+    const selectedProfileImage: ReactNode = (
+      <>
+        <img
+          className="profile-upload__image profile-upload__image--selected"
+          src={selectedImageUrl}
+          alt="Profile Upload Image"
```
LilBank: How about **Uploaded profile image**?

src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+        />
+        <img
+          className="profile-upload__icon"
+          src="/icons/pencil.svg"
+          alt="Profile Upload Icon"
```
LilBank: How about **Profile image edit icon**>

src/styles/components/_profile-upload.scss
```diff
@@ -0,0 +1,108 @@
...
+      bottom: rem(14px);
+      right: rem(14px);
+
+      width: rem(24px);
+      height: rem(24px);
+    }
+  }
+
+  &__caption {
+    color: #333333;
```
Kittinske15: please use colour variable that you have already assign.
```suggestion
    color: $dark-charcoal;
```

src/styles/components/_profile-upload.scss
```diff
@@ -0,0 +1,108 @@
...
+    @include respond-above('lg') {
+      font-size: rem(12px);
+    }
+
+    @include respond-above('xl') {
+      font-size: rem(14px);
+    }
+
+    & > span:first-child {
+      margin-right: rem(24px);
```
Kittinske15: I think this one no need to specify as ```first-child``` since you are using only one ``span``
```suggestion
    & > span {
      margin-right: rem(24px);
```

Kittinske15 (changes requested):

src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+          onChange={handleInputChange}
+        />
+        {renderProfileImage()}
+      </label>
+    )
+  }
+
+  return (
+    <figure className="profile-upload">
+      {renderProfileUpload()}
```
LilBank: The function `renderProfileUpload` depends too much on the function `renderProfileImage`, so as both functions depends directly on the `selectedImageUrl`. This is hard for future refactoring and improvement. How about making this function accepts 2 parameters which are the `imageUrl` and `ReactNode`?
```suggestion
      {renderProfileUpload(selectedImageUrl, renderProfileImage(selectedImageUrl)}
```

src/styles/abstracts/_variables.scss
```diff
@@ -18,6 +18,8 @@ $tornado: #868686;
 $coral-red: #ff3c3c;
 $dove-gray: #646464;
 $azure-radiance: #0297fd;
+$mercury: #e1e1e1;
+$dark-charcoal: #333333;
```
LilBank: This should also work fine.
```suggestion
$dark-charcoal: #333;
```

src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+import useTranslation from 'next-translate/useTranslation'
+import Trans from 'next-translate/Trans'
+
+const ProfileUpload = () => {
+  const [selectedImageUrl, setSelectedImageUrl] = useState<string>('')
+  const router = useRouter()
+  const { t } = useTranslation('sign-up')
+
+  useEffect(() => {
+    // Trigger when refreshing page
```
LilBank: Please remove the comments.

src/styles/components/_profile-upload.scss
```diff
@@ -0,0 +1,108 @@
+@import '../abstracts/_abstracts-dir';
+
+.profile-upload {
+  flex-direction: column;
+  box-sizing: border-box;
```
LilBank: Please use paddings to limit element content.

Screen Shot 2564-03-15 at 15 31 23

src/styles/components/_profile-upload.scss
```diff
@@ -0,0 +1,108 @@
...
+      }
+    }
+  }
+
+  &__input {
+    display: none;
+    visibility: hidden;
+  }
+
+  &__image {
```
LilBank, line 39: Try see the size of the default image and after uploaded, it wiggles.

### Before

Screen Shot 2564-03-15 at 15 31 11

### After

Screen Shot 2564-03-15 at 15 31 46

LilBank (changes requested):

src/components/ProfileUpload/index.tsx
kimvcp: Per our last conversation, users have to upload their profile image at the registration because we have to validate their profile before making any transactions.

src/components/ProfileUpload/index.tsx
kimvcp: Good catch! thanks for good comment. I really love `SOLID` principle.

src/styles/components/_profile-upload.scss
kimvcp: Thanks for mentioning this, BTW, this margin is the default value from `figure` tag.

src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+import Trans from 'next-translate/Trans'
+
+const ProfileUpload = () => {
+  const [selectedImageUrl, setSelectedImageUrl] = useState<string>('')
+  const router = useRouter()
+  const { t } = useTranslation('sign-up')
+
+  useEffect(() => {
+    const handleWindowClose = (event: BeforeUnloadEvent) => {
+      if (!selectedImageUrl) return
```
LilBank: Please do not return `undefined`. It is an ineffective way of error handling. Try returning something useful of go with `try...catch`

src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+      return (event.returnValue = t('warningText'))
+    }
+
+    const handleBrowseAway = () => {
+      if (!selectedImageUrl) return
+      const isConfirmed: boolean = window.confirm(t('warningText'))
+
+      if (!isConfirmed) {
+        router.events.emit('routeChangeError')
+        throw 'routeChange aborted.'
```
LilBank, line 22: A `throw` can break the application, better to use `try...catch`.

src/components/ProfileUpload/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }, [selectedImageUrl])
+
+  const handleInputChange = (setProfileImage: (image: string) => void) => (
+    event: React.ChangeEvent<HTMLInputElement>
+  ) => {
+    const { files } = event.target
+    const hasSingleFile: boolean = files?.length === 1
+
+    if (files && hasSingleFile) {
+      const imageUrl: string = URL.createObjectURL(files[0])
```
LilBank: What do you think of avoiding the magic value? `files[0]` can be confusing sometimes.

src/styles/components/_profile-upload.scss
LilBank: `box-sizing: border-box` has no effect on margin. Please recheck.

src/styles/components/_profile-upload.scss
LilBank: Screen Shot 2564-03-17 at 14 57 43
Screen Shot 2564-03-17 at 14 57 57

They still wiggles 🤔

LilBank (changes requested):

src/components/ProfileUpload/index.tsx
kimvcp: This `throw` is used to stop navigating. Plus, it will be safety ignored.
Further reading [here](https://github.com/vercel/next.js/discussions/12348#discussioncomment-8089)

LilBank (approved):

Kittinske15 (approved):

PR #46, opened 2021-03-12
[ch124] As a user, I can see Goal section in About us screen

Kittinske15:
https://app.clubhouse.io/cobda/story/124/as-a-user-i-can-see-goal-section-in-about-us-screen

## Why?

This section appends connectedly under the introduction section. It shows our goal section.

## Changes

- Our goal section consists of ``header``, ``title`` and ``description``

## Screenshot(s)

**Before**

image

**After**

image


src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,46 @@
...
+    line-height: 1.53;
+
+    @include respond-above('md') {
+      height: auto;
+      padding: 0 rem(28px 20px 28px);
+      font-size: rem(16px);
+    }
+
+    @include respond-above('lg') {
+      padding: 0 rem(180px 56px 80px);
```
kimvcp: I don't think this is a good idea to make this huge amount of right padding (180px). Should we keep the description center like our design in `desktop-1920` instead?

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  &__description {
+    box-sizing: border-box;
+    padding: rem(16px 28px);
+    margin: auto;
+    font-size: rem(14px);
+    font-weight: 200;
+    text-align: center;
+    line-height: 1.53;
```
kimvcp: CSS ordering please
```suggestion
    margin: auto;
    
    font-size: rem(14px);
    font-weight: 200;
    line-height: 1.53;
    text-align: center;

```

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,46 @@
...
+  border: solid 1px $silver-chalice;
+  background-color: transparent;
+
+  &__title {
+    display: none;
+
+    @include respond-above('md') {
+      display: block;
+      padding: rem(32px) 0 rem(16px);
+      margin: 0;
```
kimvcp: Please add empty space between `Text` and `Box model`

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,46 @@
+@import '../abstracts/_abstracts-dir';
+
+.our-goal-box {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  border: solid 1px $silver-chalice;
+  background-color: transparent;
```
kimvcp: Since `transparent` is the default value of `background-color`, Do we really need to override this?

kimvcp (changes requested):

src/styles/components/_our-goal-box.scss
Kittinske15: No need krub, my bad.

src/styles/components/_our-goal-box.scss
Kittinske15: Sure, at first i'm keeping this with design in ``desktop-1440``

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,48 @@
+@import '../abstracts/_abstracts-dir';
+
+.our-goal-box {
```
LilBank: `our-goal` is already a block element. I don't think we need the prefix `box`.

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,48 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  border: solid 1px $silver-chalice;
+
```
LilBank: What about adding the `padding-top/bottom` at the block level instead of adding at `h` and `p`?

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,48 @@
...
+  &__title {
+    display: none;
+
+    @include respond-above('md') {
+      display: block;
+      padding: rem(32px) 0 rem(16px);
```
LilBank: How about `margin-bottom: 16px`?

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,48 @@
...
+    text-align: center;
+
+    @include respond-above('md') {
+      height: auto;
+      padding: 0 rem(28px 32px 28px);
+      font-size: rem(16px);
+    }
+
+    @include respond-above('lg') {
+      padding: 0 rem(110px 44px 110px);
```
LilBank: This could also be written like this `padding: 0 rem(110px 44px) => (top left-right bottom)`

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,48 @@
...
+    @include respond-above('lg') {
+      padding: rem(44px) 0 rem(24px);
+      font-size: rem(36px);
+    }
+  }
+
+  &__description {
+    box-sizing: border-box;
+    padding: rem(16px 28px);
+    margin: auto;
```
LilBank: This could be set to 0 though.

src/styles/components/_our-goal-box.scss
```diff
@@ -0,0 +1,48 @@
...
+
+    font-size: rem(14px);
+    font-weight: 200;
+    line-height: 1.53;
+    text-align: center;
+
+    @include respond-above('md') {
+      height: auto;
```
LilBank: Please recheck the usage of this.

src/styles/components/_about-us-section.scss
```diff
@@ -20,11 +20,16 @@
     display: flex;
     justify-content: center;
 
-    &--quote {
+    &--quote,
+    &--goal {
```
LilBank: No need to duplicate this, how about creating a single modifier which the name fits the action? (The action is to hide in small screen, so think of a name for it instead of having two modifiers doing the same thing. **Modifier** is to modify => modify things differently.)

LilBank (changes requested):

LilBank (approved):

kimvcp (approved):

Raksani (approved):

PR #49, opened 2021-03-23
[ch224] As a user, I can see Product Information section in Product screen

kimvcp:
https://app.clubhouse.io/cobda/story/224/as-a-user-i-can-see-product-information-section-in-product-screen

## Why?

This **PR** is for adding Product Information which appends connectedly under the Product Detail section.

## Changes

- Product Information is containing the `title`, `subtitle`, and `description`.

### Before

Screen Shot 2564-03-24 at 01 19 26

### After

Mobile

Screen Shot 2564-03-24 at 01 14 17

Desktop

Screen Shot 2564-03-24 at 01 14 02


Kittinske15 (approved):

src/styles/layouts/_layout-product-view.scss
```diff
@@ -18,7 +18,13 @@
   }
 
   @include respond-above('xl') {
+    grid-template-columns: 1fr minmax(rem(1104px), rem(1440px)) 1fr;
```
LilBank: Let's create a variable for these.

src/styles/components/_product-information.scss
```diff
@@ -1,20 +1,122 @@
...
+    font-size: rem(14px);
+    line-height: 1.3;
+
+    @include respond-above('sm') {
+      font-size: rem(16px);
+    }
+
+    @include respond-above('md') {
+      color: $darker-gunpowder-gray;
+      font-size: rem(17px);
```
LilBank: I think we could just stick with `16px`.

src/styles/layouts/_layout-product-view.scss
```diff
@@ -31,15 +37,21 @@
 
   & > main {
     grid-area: main;
-    
-    padding: rem(24px 16px);
```
LilBank, line 35: There's no padding in mobile screen, left and right should be `16px`.

src/styles/components/_product-content.scss
```diff
@@ -7,14 +7,31 @@
...
-  @include respond-above('lg') {
+  @include respond-above('md') {
     grid-template-columns: repeat(2, 1fr);
     grid-template-rows: repeat(2, 1fr);
     grid-template-areas:
       'media detail'
       'information information';
   }
 
+  & > div {
```
LilBank: Since below this class we specify the selected class explicitly (ex: `& > .product-...`), this should also be specify explicitly too. For readability and also the block we're selecting is a sub-main (next level from `main`) block which is not a small element like simple `div` with one level `span` (`div > span`).

src/styles/components/_product-information.scss
```diff
@@ -1,20 +1,122 @@
...
+    }
+
+    @include respond-above('lg') {
+      margin-bottom: rem(24px);
+    }
   }
 
   &__title {
-    margin: 0 0 rem(16px) 0;
+    all: unset;
```
LilBank, line 46: Quick question, if this is just to reset the `margin`, does it necessary to unset **all**?
PS. Even `description` doesn't have this.

src/styles/components/_product-information.scss
```diff
@@ -1,20 +1,122 @@
...
+    }
+
+    @include respond-above('lg') {
+      width: rem(220px);
+      padding: rem(24px) 0 0 rem(32px);
+    }
+  }
+
+  &__subheader {
+    display: none;
```
LilBank, line 33: Quick question, so this is intentionally hidden?

LilBank (changes requested):

src/styles/components/_product-information.scss
kimvcp: Then it will be the same size as the title (`Product Details`). As another option, do you think we should make the title bigger than description? See below.

Same size (please ignore the subheader)

Screen Shot 2564-03-27 at 23 48 09

Title is bigger

Screen Shot 2564-03-27 at 23 47 47

src/styles/layouts/_layout-product-view.scss
kimvcp: Do you mean to add `padding-left` and `padding-right` here? 
The output would be:

`padding` added
Screen Shot 2564-03-27 at 23 55 44

Our design reference (no `padding`)
Screen Shot 2564-03-27 at 23 54 29

src/styles/components/_product-information.scss
kimvcp: Not just `margin` but also `font-weight` since `h2` is `bold` by default.

With the overridden `margin` and `font-weight`
Screen Shot 2564-03-28 at 00 02 30

Do you think this one is cleaner?
Screen Shot 2564-03-28 at 00 01 46

src/styles/components/_product-information.scss
kimvcp: Yep there is no `subheader` in mobile. References are the screenshots in PR description.

src/styles/layouts/_layout-product-view.scss
kimvcp: Good idea

src/styles/components/_product-content.scss
kimvcp: You mean to change it to this below. 

Screen Shot 2564-03-28 at 00 11 14

Or this
Screen Shot 2564-03-28 at 00 13 13

If there are better ways of writing this please suggest.

src/styles/components/_product-information.scss
LilBank: I cheer the 2nd one.

src/styles/layouts/_layout-product-view.scss
LilBank: No padding is a mis-design. Phones with curved edge of little bezel will encounter the mis-touch problem.

src/styles/components/_product-content.scss
LilBank: The first one is the cleanest and readable for now. 😎

src/styles/components/_product-information.scss
LilBank: Agree with using unset.

locales/en/product-view.json
```diff
@@ -0,0 +1,3 @@
+{
+  "productDetail": "Product Details"
```
Raksani: I was wondering if we can use the same word "detail" without 's' ☺️ Btw, in terms of grammar, both are correct.

LilBank (approved):

PR #56, opened 2021-04-21
[ch230] As a user, I can see Product Media section in Product View screen

Raksani:
https://app.clubhouse.io/cobda/story/230/as-a-user-i-can-see-product-media-section-in-product-view-screen

## Why?

To show the product's image(s) of the sale post which be able to see a large image.

## Changes

- Background of the product media section
- Having Product's image has shown in large size
- Having an Image selector to select and browse more images of certain product

## Screenshot(s)
**Before**
Screenshot 2564-04-21 at 16 59 01

**After**
**Desktop version**
Screenshot 2564-04-21 at 16 47 14

**Mobile version**
Screenshot 2564-04-21 at 16 46 52

### DIFF
--- a/src/components/ProductContent/ProductMedia.tsx
+++ b/src/components/ProductContent/ProductMedia.tsx
@@ -1,54 +1,47 @@
 import React from 'react'
-import { useState, useEffect } from 'react'
-import cx from 'classnames'
+import { useState } from 'react'
 
-//TO DO: Retrieve images from database instead
+//TODO: Retrieve images from database instead
 const imagePaths: string[] = [
   '/images/nike-jordan-off-white.jpg',
   '/images/nike-woman-running-shoe.jpg',
-  '/images/nike-air-vapormax-360-herren.jpg'
+  '/images/yeezy-380.jpg'
 ]
 
-const useImage = ({ src }: any) => {
-  const [isLoaded, setLoaded] = useState(false)
-
-  useEffect(() => {
-    const img = new Image()
-    img.src = src
-    img.onload = () => setLoaded(true)
-  }, [src])
-  return { isLoaded }
-}
-
-const Cover = ({ src }: any) => {
-  const { isLoaded } = useImage({ src })
-  return <img className={cx('product-media__image', 'smooth', { isLoaded })} src={src} />
-}
-
-const useSrc = () => {
-  const [selectedSrc, setSrc] = useState(imagePaths[0])
-  return {
-    imageSelectorButton: imagePaths.map((src) => (
-      <button
-        // TO DO: Change image location
-        style={{ background: `url(${src}) no-repeat; background-size: cover` }}
-        className={cx('product-media__button', { active: src === selectedSrc })}
-        onClick={() => setSrc(src)}
-        key={src}
-      ></button>
-    )),
-    selectedImageCover: imagePaths.map((src) => (src === selectedSrc ? <Cover key={src} src={selectedSrc} /> : null))
-  }
-}
-
 const ProductMedia = () => {
-  const { imageSelectorButton, selectedImageCover } = useSrc()
+  const [selectedImagePath, setSelectedImagePath] = useState<string>(imagePaths[0])
+
+  const renderImage = () => (
+    <div className="product-media__image-wrapper">
+      <img className="product-media__image" src={selectedImagePath} />
+    </div>
+  )
+
+  const renderImageSelector = () => {
+    const images = imagePaths.map((path, index) => {
+      const handleImageClick = (path: string) => () => {
+        setSelectedImagePath(path)
+      }
+
+      const imageClassName: string =
+        path === selectedImagePath
+          ? 'product-media__thumbnail product-media__thumbnail--active'
+          : 'product-media__thumbnail'
+
+      return (
+        <div key={index} className="product-media__thumbnail-wrapper" onClick={handleImageClick(path)}>
+          <img className={imageClassName} src={path} alt="Product Thumbnail" />
+        </div>
+      )
+    })
+
+    return <div className="product-media__image-selector">{images}</div>
+  }
 
   return (
     <div className="product-media">
-      <div className="product-media__image-container">{selectedImageCover}</div>
-      <br />
-      <div className="product-media__image-selector">{imageSelectorButton}</div>
+      {renderImage()}
+      {renderImageSelector()}
     </div>
   )
 }
--- a/src/components/ProductContent/ProductMedia.tsx
+++ b/src/components/ProductContent/ProductMedia.tsx
@@ -1,4 +1,58 @@
 import React from 'react'
+import { useState, useEffect } from 'react'
+import cx from 'classnames'
+
+//TO DO: Retrieve images from database instead
+const srcs = [
+  '/images/nike-jordan-off-white.jpg',
+  '/images/nike-woman-running-shoe.jpg',
+  '/images/nike-air-vapormax-360-herren.jpg',
+]
+
+interface ISrc {
+  src: string
+}
+
+const useImage = ({ src }: ISrc) => {
+  const [loaded, setLoaded] = useState(false)
+
+  useEffect(() => {
+    const img = new Image()
+    img.src = src
+    img.onload = () => setLoaded(true)
+  }, [src])
+  return {
+    loaded,
+  }
+}
+
+const Cover = ({ src }: ISrc) => {
+  const { loaded } = useImage({ src })
+  return (
+    <img
+      className={cx('product-media__image', 'smooth', { loaded })}
+      src={src}
+    />
+  )
+}
+
+const useSrc = () => {
+  const [selectedSrc, setSrc] = useState(srcs[0])
+  return {
+    buttons: srcs.map((src) => (
+      <button
+        // TO DO: Change image location
+        style={{ background: `url(${src}) no-repeat; background-size: cover;` }}
+        className={cx('product-media__button', { active: src === selectedSrc })}
+        onClick={() => setSrc(src)}
+        key={src}
+      ></button>
+    )),
+    cover: srcs.map((src) =>
+      src === selectedSrc ? <Cover key={src} src={selectedSrc} /> : null
+    ),
+  }
+}
 
 const ProductMedia = () => {
   return (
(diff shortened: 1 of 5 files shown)


src/styles/components/_product-media.scss
```diff
@@ -1,30 +1,86 @@
...
+
+    background: $light;
+  }
+
+  @include respond-above('lg') {
+    height: 560px;
+  }
+
+  @include respond-above('xl') {
+    height: 650px;
```
Kittinske15: Don't forget to put ``rem()``. Fix other places too.
```suggestion
  @include respond-above('md') {
    height: rem(430px);
    border-radius: 15%;

    background: $light;
  }

  @include respond-above('lg') {
    height: rem(560px);
  }

  @include respond-above('xl') {
    height: rem(650px);
```

src/styles/components/_product-content.scss
```diff
@@ -1,43 +1,75 @@
...
     @include respond-above('lg') {
-      margin-bottom: rem(48px);
+      margin-bottom: rem(30px);
     }
   }
 
+  & > .product-content__media,
+  & > .product-content__detail {
+    margin-top: rem(16px);
+  }
```
Kittinske15:
```suggestion
  &__media,
  &__detail {
    margin-top: rem(16px);
  }
```

src/styles/components/_product-media.scss
```diff
@@ -1,30 +1,86 @@
...
-    width: 60%;
-    height: 60%;
-    background-color: blueviolet;
+    display: block;
+    max-width: 90%;
+    max-height: 100%;
+  }
+
+  &__image.smooth {
+    @include transition($what: opacity, $time: 0.2s, $how: cubic-bezier(0.3, 0, 0.4, 1));
```
Kittinske15:
```suggestion
    @include transition(opacity, 0.2s, cubic-bezier(0.3, 0, 0.4, 1));
```

src/styles/components/_product-media.scss
```diff
@@ -1,30 +1,86 @@
...
+    }
+
+    @include respond-above('lg') {
+      height: 420px;
+    }
+
+    @include respond-above('xl') {
+      height: 500px;
+    }
   }
```
Kittinske15, line 39: Better set ``margin-right`` in this element.
```
  &__image-selector {
    & > *:not(:last-child) {
      margin-right: rem(5px);
    }
  }
```

src/styles/components/_product-media.scss
```diff
@@ -1,30 +1,86 @@
...
+  &__image.smooth.loaded {
+    opacity: 1;
+  }
+
+  &__button {
+    width: rem(80px);
+    height: rem(80px);
+    border: none;
+    border-radius: 10%;
+    margin-right: rem(5px);
```
Kittinske15: See this [comment](https://github.com/Cobda/cobda-web/pull/56#discussion_r619648306)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
kimvcp: This file should be removed

package-lock.json
```diff
@@ -0,0 +1,4429 @@
+{
```
kimvcp: This file should be removed. Please use `yarn add <package name>` to install stuffs instead of `npm`.

src/components/ProductContent/ProductMedia.tsx
```diff
@@ -1,16 +1,66 @@
 import React from 'react'
+import { useState, useEffect } from 'react'
+import cx from 'classnames'
+
+//TO DO: Retrieve images from database instead
```
kimvcp: Normally, we don't add space here. You can simply select and change all occurrences.
```suggestion
//TODO: Retrieve images from database instead
```

src/components/ProductContent/ProductMedia.tsx
```diff
@@ -1,16 +1,66 @@
...
+const srcs = [
+  '/images/nike-jordan-off-white.jpg',
+  '/images/nike-woman-running-shoe.jpg',
+  '/images/nike-air-vapormax-360-herren.jpg',
+]
+
+interface ISrc {
+  src: string
+}
```
kimvcp: 3 things here:
1. We have agreed to name the interface without prefix `I`.
2. Seems like we don't really need to have an interface here, a simple approach will do the job.
3. Do you think `imagePath` is more meaningful than `src` ?

```suggestion
const imagePaths: string = [
  '/images/nike-jordan-off-white.jpg',
  '/images/nike-woman-running-shoe.jpg',
  '/images/nike-air-vapormax-360-herren.jpg',
]
```

src/components/ProductContent/ProductMedia.tsx
```diff
@@ -1,16 +1,66 @@
...
+        className={cx('product-media__button', { active: src === selectedSrc })}
+        onClick={() => setSrc(src)}
+        key={src}
+      ></button>
+    )),
+    cover: srcs.map((src) =>
+      src === selectedSrc ? <Cover key={src} src={selectedSrc} /> : null
+    ),
+  }
+}
```
kimvcp: Correct me if I am wrong, you are trying to create a custom `hook` here right?. If so, I don't think we need to do that since we only create a custom `hook` when we want to make it reusable for other classes. Further reading [here](https://reactjs.org/docs/hooks-custom.html#using-a-custom-hook)

I have 3 concerns here:

1. We all agreed to name a boolean variable using `is/has` as a prefix. In your code, `loaded` should be `isLoaded.`
2. Since we are using BEM CSS convention, instead of using those `inline style` (` style={{ background: url...`), we could create a CSS class instead.
3. The approach you are using is good, but I think we should use the same approach as other members. This package (`import cx from 'classnames'`) is another approach which is different from ours. You could see `ProductInformation.tsx` as an example.

src/components/ProductContent/ProductMedia.tsx
```diff
@@ -1,16 +1,66 @@
...
 
 const ProductMedia = () => {
+  const { buttons, cover } = useSrc()
   return (
```
kimvcp, line 41: Always add space before `return`
```suggestion

  return (
```

src/components/ProductContent/ProductMedia.tsx
```diff
@@ -1,16 +1,66 @@
...
-        <h3 className="product-media__title">
-          Images
-        </h3>
-      </header>
-      <div className="product-media__image-container">
-        <div className="product-media__image"></div>
-      </div>
+      <div className="product-media__image-container">{cover}</div>
+      <br />
+      <div className="product-media__image-selector">{buttons}</div>
```
kimvcp: Do you think the words `cover` and `buttons` are too vague? Just specify what action or what type of the button is.
According to this book [Clean Code](https://enos.itcollege.ee/~jpoial/oop/naited/Clean%20Code.pdf) page 39, If we name variable descriptively, others member can take less time to understand your code.

src/styles/components/__components-dir.scss
```diff
@@ -38,7 +38,6 @@
 @import './profile-card';
 @import './carousel';
 @import './carousel-item';
-@import 'pure-react-carousel/dist/react-carousel.es.css';
```
kimvcp: Did you accidentally remove this?

src/styles/components/_product-media.scss
```diff
@@ -1,30 +1,86 @@
 .product-media {
-  display: flex;
+  @include flex-center();
```
kimvcp: Please group `@include` in the same place
```
  flex-direction: column;
  box-sizing: border-box;
  height: 450px;
  padding: rem(16px);
  border-radius: 15%;

  @include flex-center();

  @include respond-above('sm') {
    height: 600px;
  }
```

src/styles/components/_product-media.scss
```diff
@@ -1,30 +1,86 @@
...
   }
 
-  &__title {
-    margin: 0;
+  @include respond-above('md') {
+    height: 430px;
+
+    border-radius: 15%;
+
+    background: $light;
```
kimvcp: How about adding space only when there are more than 1 css properties for each group? 

Do you think this looks cleaner
```suggestion
    height: 430px;
    border-radius: 15%;
    background: $light;
```

src/styles/components/_product-media.scss
```diff
@@ -1,30 +1,86 @@
...
   &__image-container {
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    width: rem(200px);
-    height: rem(200px);
-    background-color: black;
+    @include flex-center();
+
+    height: 500px;
```
kimvcp: Same [here](https://github.com/Cobda/cobda-web/pull/56#discussion_r619662584)
```suggestion
    height: 500px;
    
    @include flex-center();
```

kimvcp (changes requested):

src/styles/components/_product-content.scss
Raksani: From the original file we use `& > .product-content__media` and to change it to be `&__media` is it change anything? Just asking so I can implement it in further tasks. Thanks.

src/styles/components/_product-media.scss
Raksani: I'm sorry. I can't see the comment. Please mention it again.

src/components/ProductContent/ProductMedia.tsx
Raksani: I agree with the naming. But in `useImage` and `Cover` I encountered the error `Binding element 'src' implicitly has an 'any' type.` so that I use interface to fix this problem. Do you mind providing suggestions for this issue? or Can I change it to be `any` type instead? like this `const useImage = ({ src }: any) => {`


And do you mean `const imagePaths: string[] = [...]` right?

src/components/ProductContent/ProductMedia.tsx
Raksani: Alright. I'll fix it.

src/styles/components/_product-content.scss
Kittinske15: It will change nothing because the ``&`` is a nest styles for the element and modifier within the same parent declaration block which this one the ``&`` has a meaning of its class name ``.product-content`` already.

Kittinske15 (approved):

